PR #7, opened 2019-01-21
Type Inference Task

aalexandrasilva:
The following features have been implemented:
1) Inference of the type of the subjects and objects of the model through the domain and range of a property in the ontology
2) Resources without type are being counted before and after this preprocessing


src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -0,0 +1,146 @@
...
+import org.apache.jena.rdf.model.Property;
+import org.apache.jena.rdf.model.RDFNode;
+import org.apache.jena.rdf.model.Resource;
+import org.apache.jena.rdf.model.ResourceFactory;
+import org.apache.jena.rdf.model.Statement;
+import org.apache.jena.util.FileManager;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class Inferer {
```
MichaelRoeder, line 32: Could you please add a javadoc comment to classes (and in the best case even to methods) that explains the class (or method)?
Note that comments for getter and setter methods are not necessary :wink:

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -0,0 +1,146 @@
...
+
+	private static final Logger LOGGER = LoggerFactory.getLogger(Inferer.class);
+	private int counter;
```
MichaelRoeder: This counter is used as an attribute although I do not really see a reason to have it as a class attribute.

It is ok, if you want to count something to log the number after counting (as you do it in the `checkEmptyTypes` method). However, you directly reset the counter after using it, i.e., you could also simply use a local counter variable which makes it much less error prone.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -0,0 +1,146 @@
...
+
+	public Inferer() {
+		this.counter = 0;
+	}
+
+	public Model process(Model personModel, Map<String, String> map, String fileName, String dataFolderPath) {
```
MichaelRoeder: `personModel` is very closely bound to the use case that you had. However, it should be a more general name, e.g., `originalModel` or `sourceModel` since we will use it for other models as well :wink: 

`map` is a very generic name. You may try to find a name which expresses its meaning.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -0,0 +1,146 @@
...
+			extractProperties(newModel, ontModel);
+			checkEmptyTypes(set, newModel);
+		}
+		return newModel;
+	}
+
+	private Set<Resource> extractUniqueResources(Model newModel) {
+		Set<Resource> set = new HashSet<>();
+		List<Statement> statements = newModel.listStatements().toList();
+		Property type = ResourceFactory.createProperty("https://www.w3.org/1999/02/22-rdf-syntax-ns#type");
```
MichaelRoeder: You do not have to create a new property. For "well-known" properties, there are predefined classes, e.g., `RDF.type`.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -0,0 +1,146 @@
...
+		if (ontModel != null) {
+			extractProperties(newModel, ontModel);
+			checkEmptyTypes(set, newModel);
+		}
+		return newModel;
+	}
+
+	private Set<Resource> extractUniqueResources(Model newModel) {
+		Set<Resource> set = new HashSet<>();
+		List<Statement> statements = newModel.listStatements().toList();
```
MichaelRoeder: Creating this list is an additional overhead... :thinking: 
(although the overhead is not huge, but still :wink: )
At the moment, you are creating an iterator, let it iterate over all statements to write them into a list to iterate over this list, again.

It might be better to simply use the iterator that you already got from the `listStatements()` method :wink:

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -0,0 +1,146 @@
...
+			LOGGER.info(e.getMessage());
+		}
+		if (ontModel != null) {
+			extractProperties(newModel, ontModel);
+			checkEmptyTypes(set, newModel);
+		}
+		return newModel;
+	}
+
+	private Set<Resource> extractUniqueResources(Model newModel) {
```
MichaelRoeder: The method is not exactly doing what the name `extractUniqueResources` promises. You are extracting unique resources that have no type information, right?

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -0,0 +1,146 @@
...
+				Statement objType = ResourceFactory.createStatement(object.asResource(), type, range.get(i));
+				if (!model.contains(objType)) {
+					stList.add(objType);
+				}
+			}
+		}
+		return stList;
+	}
+
+	public Map<String, String> mapModel2Ontology() {
```
MichaelRoeder: This bounds your implementation very closely to the person graph example. This shouldn't be done. Please enable the class to take the file -> ontology map as input instead of generating it here.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	public OntModel readOntology(Map<String, String> map, String fileName, String dataFolderPath) throws IOException {
+		OntModel ontModel = ModelFactory.createOntologyModel(OntModelSpec.OWL_MEM);
+		String fullPath = null;
+		//if no path is provided, it's a test
+		if(dataFolderPath == null) {
+			fullPath = map.get(fileName);
+		} else {
+			fullPath = dataFolderPath + "\\ontologies\\" + map.get(fileName);
```
MichaelRoeder: This is a little bit strange. You are forcing the user to have `\ontologies\` in its path (which is also only working on Windows OS). You should offer the user to give you a single `File` (or a String) object and simply use load it without any additional changes. The user who is calling your Inferer will be responsible for giving you the correct paths.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -0,0 +1,146 @@
...
+		}
+		InputStream inputStream = FileManager.get().open(fullPath);
```
MichaelRoeder: I think that this can be handled in a better way, e.g.,
```java
try(InputStream inputStream = FileManager.get().open(fullPath)) {
    ontModel.read(inputStream, "RDF/XML");
} catch (IOException e) {
    LOGGER.error("Couldn't read ontology file. Returning empty ontology model.",e);
}
```
This has the advantage that the input stream is still closed even if the `read` methods throws an exception. However, maybe the check whether the inputStream is null is still necessary. Or you simply use:
```java
try(InputStream inputStream = new BufferedInputStream(new FileInputStream(fullPath)))
```

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -0,0 +1,146 @@
...
+	public void extractProperties(Model model, OntModel ontModel) {
+		List<Statement> curModelList = model.listStatements().toList();
+		for (Statement statement : curModelList) {
+			List<Statement> statements = searchType(statement, ontModel, model);
+			model.add(statements);
+		}
+	}
+
+	private List<Statement> searchType(Statement statement, OntModel ontModel, Model model) {
+		List<Statement> stList = new ArrayList<>();
```
MichaelRoeder: You may want to use `Set<Statement> newStmts = new HashSet<>();` instead.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -0,0 +1,146 @@
...
+			for (int i = 0; i < domain.size(); i++) {
+				Statement subjType = ResourceFactory.createStatement(subject, type, domain.get(i));
+				if (!model.contains(subjType)) {
+					stList.add(subjType);
+				}
+			}
+			
+			List<? extends OntResource> range = temp.listRange().toList();
+			for (int i = 0; i < range.size(); i++) {
+				Statement objType = ResourceFactory.createStatement(object.asResource(), type, range.get(i));
```
MichaelRoeder: Please note that your object does not have to be a resource. Therefore, you should check it with `.isResource()` before transforming it into a resource. Otherwise, your method will crash.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -0,0 +1,146 @@
...
+		Resource subject = statement.getSubject();
+		Property predicate = statement.getPredicate();
+		RDFNode object = statement.getObject();
+		Property type = ResourceFactory.createProperty("https://www.w3.org/1999/02/22-rdf-syntax-ns#type");
+
+		// search for the predicate of the model in the ontology
+		OntProperty temp = ontModel.getOntProperty(predicate.toString());
+		if (temp != null) {
+			List<? extends OntResource> domain = temp.listDomain().toList();
+			for (int i = 0; i < domain.size(); i++) {
```
MichaelRoeder: Maybe the `for(OntResource d : domain)` loop would be better since you do not know which list implementation is used in the back and whether `.get(i)` is fast or slow. Since you are not interested in the id (`i`) of the element and just want to iterate over all existing domains, this would do the job in a better way.

src/test/resources/test.ttl
```diff
@@ -0,0 +1,10 @@
...
+<http://dbpedia.org/resource/Hardin_County,_Kentucky> <https://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.w3.org/2002/07/owl#Thing>  .
+<http://dbpedia.org/resource/Abraham_Lincoln> <http://dbpedia.org/ontology/birthPlace> <http://dbpedia.org/resource/Kentucky>  .
+<http://dbpedia.org/resource/Kentucky> <https://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.w3.org/2002/07/owl#Thing>  .
+<http://dbpedia.org/resource/Aristotle> <http://dbpedia.org/ontology/deathPlace> <http://dbpedia.org/resource/Chalcis>  .
+<http://dbpedia.org/resource/Chalcis> <https://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://dbpedia.org/ontology/Location>  .
+<http://dbpedia.org/resource/Chalcis> <https://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://dbpedia.org/ontology/Place>  .
+<http://dbpedia.org/resource/Chalcis> <https://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://schema.org/Place>  .
+<http://dbpedia.org/resource/Chalcis> <https://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.wikidata.org/entity/Q3455524>  .
+<http://dbpedia.org/resource/Chalcis> <https://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.wikidata.org/entity/Q486972>  .
```
MichaelRoeder: the test already looks good. However, could you please add some triples with literals? I think that would show you that your current implementation can not handle literals correctly.

MichaelRoeder (changes requested):
The solution already looks good. However, some changes are needed.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
aalexandrasilva: The idea behind this method was to extract all the resources in the model with the exception of the ones that were in a type statement. I was using this set purely to do the before and after count of the resources that had no type information. 
I have changed this method to always add the subject but to add the object only if a) it's a resource and b) it's not part of a type statement.

aalexandrasilva:
@MichaelRoeder The requested changes have been implemented.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
MichaelRoeder: I see. However, I am not sure whether point b) is necessary. Imagine a class that has a label or some other, additional triple. You would exclude it when it is the object in a triple but still add it to your list, if it is the subject in the triple defining the label. Apart from that, I do not see a reason why class resources should be excluded.

src/test/resources/test.ttl
MichaelRoeder: What about the comment above?

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
aalexandrasilva: Oh I understand now! I have removed that condition.

src/test/resources/test.ttl
aalexandrasilva: It was indeed crashing before the changes. I have added a new test case with literals in the last commit.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -0,0 +1,178 @@
...
+	 * property in the ontology and adding the inferred triples to the new model
+	 * @param newModel model where we will add the new triples
+	 * @param sourceModel provided model where we iterate through the statements
+	 * @param ontModel the ontology model
+	 */
+	public void iterateStmts(Model newModel, Model sourceModel, OntModel ontModel) {
+		StmtIterator stmts = sourceModel.listStatements();
+		while(stmts.hasNext()) {
+			Statement curStatement = stmts.next();
+			List<Statement> stmtsList = new ArrayList<>();
```
MichaelRoeder: I think this list is not necessary, is it?

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -0,0 +1,178 @@
...
+			for(OntResource curResource: domain){
+				Statement subjType = ResourceFactory.createStatement(subject, RDF.type, curResource);
+				if (!newModel.contains(subjType)) {
+					newStmts.add(subjType);
+				}
+			}
+			
+			List<? extends OntResource> range = temp.listRange().toList();
+			for(OntResource curResource: range){
+				if(object.isResource()) {
```
MichaelRoeder: It would make sense to move this `if` statement a little bit more "up". I think it should be already checked before asking for the range of the property. If the object is not a resource, we are not interested in the range, anyway :wink:

src/main/java/org/aksw/simba/lemming/creation/PersonGraphDataset.java
```diff
@@ -22,23 +23,31 @@ public PersonGraphDataset() {
...
 					 personModel.read(file.getAbsolutePath(), "TTL");
-					 LOGGER.info("Read data to model - "+ personModel.size() + " triples");
+					 LOGGER.info("Read data to model - "+ personModel.size() + " triples");			 
+					 
+					 Inferer inferer = new Inferer();
+					 String ontFilePath = "dbpedia_2015-04.owl";
+					 OntModel ontModel = inferer.readOntology(ontFilePath);
+					//returns a new model with the added triples
+					 //Model newModel = 
+					 inferer.process(personModel, ontModel);
```
MichaelRoeder: Shouldn't this line be `personModel = inferer.process(personModel, ontModel);` ?

src/main/java/org/aksw/simba/lemming/creation/PersonGraphDataset.java
```diff
@@ -22,23 +23,31 @@ public PersonGraphDataset() {
...
 				 
 				 if(file != null && file.isFile() && file.getTotalSpace() > 0){
 					 Model personModel = ModelFactory.createDefaultModel();
 					 //read file to model
 					 personModel.read(file.getAbsolutePath(), "TTL");
-					 LOGGER.info("Read data to model - "+ personModel.size() + " triples");
+					 LOGGER.info("Read data to model - "+ personModel.size() + " triples");			 
+					 
+					 Inferer inferer = new Inferer();
+					 String ontFilePath = "dbpedia_2015-04.owl";
```
MichaelRoeder: This is already good for testing. However, we need some way to define the mapping from file names to ontologies since the different files have different ontologies.

Additionally, it would be good to add very "general" ontologies to the `ontModel`. Especially RDF (http://www.w3.org/1999/02/22-rdf-syntax-ns#) and RDFS (http://www.w3.org/2000/01/rdf-schema#).

src/test/resources/dbpedia_test.owl
```diff
@@ -0,0 +1,2919 @@
+<rdf:RDF xml:base="http://dbpedia.org/ontology/" xmlns:foaf="http://xmlns.com/foaf/0.1/" xmlns:cc="http://creativecommons.org/ns#" xmlns:vann="http://purl.org/vocab/vann/" xmlns:dcterms="http://purl.org/dc/terms/" xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:wgs84pos="http://www.w3.org/2003/01/geo/wgs84_pos#" xmlns:cidoccrm="http://purl.org/NET/cidoc-crm/core#" xmlns:wikidata="http://www.wikidata.org/entity/" xmlns:dul="http://www.ontologydesignpatterns.org/ont/dul/DUL.owl#" xmlns:d0="http://www.ontologydesignpatterns.org/ont/d0.owl#" xmlns:owl="http://www.w3.org/2002/07/owl#" xmlns:xsd="http://www.w3.org/2001/XMLSchema#" xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#" xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#" xmlns:prov="http://www.w3.org/ns/prov#" xmlns="http://dbpedia.org/ontology/">
```
MichaelRoeder, line 1: Could you please reduce the size of this file and focus on the statements which are really necessary for the test?

MichaelRoeder (commented):
I only have some small comments left :smiley:

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -0,0 +1,177 @@
...
+			for(OntResource curResource: domain){
+				Statement subjType = ResourceFactory.createStatement(subject, RDF.type, curResource);
+				if (!newModel.contains(subjType)) {
+					newStmts.add(subjType);
+				}
+			}
+			if(object.isResource()) {
+				List<? extends OntResource> range = temp.listRange().toList();
+				for(OntResource curResource: range){
+					if(object.isResource()) {
```
MichaelRoeder: @aalexandrasilva Better than before :+1: 
However, line 146 is not needed anymore.

aalexandrasilva: yes, sorry. I did think I had deleted it.

MichaelRoeder (approved):

PR #8, opened 2019-03-14
Edge Modifier Bug Fix

aalexandrasilva:
The following features have been changed:
1) When an edge removal or addition fails, it is now returning null and assigning the error score a maximum value
2) The graph is now also being saved before the optimization process

### DIFF
--- a/src/main/java/org/aksw/simba/lemming/ColouredGraphWrapper.java
+++ b/src/main/java/org/aksw/simba/lemming/ColouredGraphWrapper.java
@@ -1,0 +1,98 @@
+package org.aksw.simba.lemming;
+
+import java.io.Serializable;
+
+import org.aksw.simba.lemming.colour.ColourPaletteWrapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.BitSet;
+import java.util.List;
+
+import grph.Grph;
+
+/**
+ * Wrapper Class to enable the persistence of the ColouredGraph class
+ * @author Ana
+ *
+ */
+public class ColouredGraphWrapper implements Serializable {
+
+	/**
+	 * 
+	 */
+	private static final long serialVersionUID = 7964247861129517487L;
+
+	private static final Logger LOGGER = LoggerFactory.getLogger(ColouredGraphWrapper.class);
+
+	protected Grph graph;
+	protected List<BitSet> vertexColours;
+	protected List<BitSet>  edgeColours;
+	protected ColourPaletteWrapper vertexPalette;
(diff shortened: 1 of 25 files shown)


src/main/java/org/aksw/simba/lemming/ColouredGraphWrapper.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.List;
+
+import grph.Grph;
+
+/**
+ * Wrapper Class to enable the persistence of the ColouredGraph class
+ * @author Ana
+ *
+ */
+public class ColouredGraphWrapper implements Serializable {
```
MichaelRoeder, line 19: I do not understand why this class is necessary. Wouldn't it be sufficient to make the ColouredGraph class implement the Serializable interface? :thinking:

src/main/java/org/aksw/simba/lemming/colour/ColourPaletteWrapper.java
```diff
@@ -0,0 +1,56 @@
...
+ *
+ */
+public class ColourPaletteWrapper implements Serializable{
+	
+	/**
+	 * 
+	 */
+	private static final long serialVersionUID = 3744241658289486040L;
+	protected Map<String, BitSet> uriColourMap = new HashMap<String, BitSet>();
+    protected int highestColourId = 0;
```
MichaelRoeder, line 20: Do we really need to keep track of this id? Isn't it automatically define by the size of the uriColourMap?

src/main/java/org/aksw/simba/lemming/colour/ColourPaletteWrapper.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.Map;
+import java.io.Serializable;
+import java.util.BitSet;
+
+/**
+ * Wrapper Class to enable the persistence of the ColourPalette class
+ * @author Ana
+ *
+ */
+public class ColourPaletteWrapper implements Serializable{
```
MichaelRoeder, line 13: Again: do we really need the wrapper class?

MichaelRoeder (commented):
It looks good. I only have some small questions.

src/main/java/org/aksw/simba/lemming/ColouredGraphWrapper.java
aalexandrasilva: At the time, I found several usages of BitSet, ObjectObjectOpenHashMap, ObjectArrayList from com.carrotsearch.hppc which do not implement Serializable. I created these Wrapper classes to use java.util classes, analogous to the hppc ones, thus allowing serialization. :)

PR #9, opened 2019-06-03
SWDF Changes

aalexandrasilva:
This pull request implements the following features:
1) The domain and range of a property and of its own equivalent properties have been condensed into a custom object. Thus, allowing us to have access to all of the domains and ranges at time of inference;
2) The equivalent properties URIs are renamed to one common URI;
3) At the end of the pre-processing, equivalent classes are renamed to a common URI.

- travis.yml was changed to openjdk10


src/main/java/org/aksw/simba/lemming/creation/Equivalent.java
```diff
@@ -0,0 +1,180 @@
...
+	 * Returns <code>true</code> if the object contains any relation to the given
+	 * element.
+	 * 
+	 * @param element given element to check for comparison
+	 * @return <code>true</code> if the set contains the element's URI or any of the
+	 *         equivalents classes/properties URI <code>false</code> otherwise
+	 */
+	@SuppressWarnings("unchecked")
+	public boolean containsElement(T element) {
+		if (((RDFNode) element).isResource() && equivalents.contains(((Resource) element).getURI()))
```
MichaelRoeder: It might be better to use `isURIResource()` instead of `isResource()` since the latter only makes sure that it is not a Literal, i.e., you can not assume that it will have a URI.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -189,6 +253,96 @@ public OntModel readOntology(String filePath) {
...
+	 * extends OntResource> class, producing a map of the Equivalent objects with
+	 * the URIs as keys.
+	 * 
+	 * @see Equivalent<T extends OntResource>
+	 * @param <T>
+	 * @param ontElements the ontology classes or properties
+	 * @return
+	 */
+	@SuppressWarnings({ "rawtypes", "unchecked" })
+	public <T extends OntResource> Map<String, Equivalent> searchEquivalents(Set ontElements) {
```
MichaelRoeder: Why you do not specify the type of the elements in the given Set? Shouldn't it be `Set<T>`?

src/main/java/org/aksw/simba/lemming/creation/Equivalent.java
```diff
@@ -0,0 +1,180 @@
...
+	 * 
+	 * @param current OntProperty / OntClass object
+	 */
+	public void addEquivalent(T current) {
+		String uri = ((Resource) current).getURI();
+		if (uri != null) {
+			this.equivalents.add(uri);
+			setName(compareNames(uri, name));
+
+			if (current instanceof OntProperty) {
```
MichaelRoeder: `current.isProperty()` would be better.

src/main/java/org/aksw/simba/lemming/creation/Equivalent.java
```diff
@@ -0,0 +1,180 @@
...
+			return true;
+
+		List<T> list = null;
+
+		try {
+			if (element instanceof OntClass) {
+				list = (List<T>) ((OntClass) element).listEquivalentClasses().toList();
+			}
+
+			if (element instanceof OntProperty) {
```
MichaelRoeder: `current.isProperty()` would be better.

src/main/java/org/aksw/simba/lemming/creation/Equivalent.java
```diff
@@ -0,0 +1,180 @@
...
+	 */
+	@SuppressWarnings("unchecked")
+	public boolean containsElement(T element) {
+		if (((RDFNode) element).isResource() && equivalents.contains(((Resource) element).getURI()))
+			return true;
+
+		List<T> list = null;
+
+		try {
+			if (element instanceof OntClass) {
```
MichaelRoeder: `current.isClass()` would be better.

src/main/java/org/aksw/simba/lemming/creation/Equivalent.java
```diff
@@ -0,0 +1,180 @@
...
+				list = (List<T>) ((OntClass) element).listEquivalentClasses().toList();
```
MichaelRoeder: `current.asClass()` would be better.

src/main/java/org/aksw/simba/lemming/creation/Equivalent.java
```diff
@@ -0,0 +1,180 @@
...
+			}
+
+			if (element instanceof OntProperty) {
+				list = (List<T>) ((OntProperty) element).listEquivalentProperties().toList();
```
MichaelRoeder: `current.asProperty()` would be better.

src/main/java/org/aksw/simba/lemming/creation/Equivalent.java
```diff
@@ -0,0 +1,180 @@
...
+	 * @param current OntProperty / OntClass object
+	 */
+	public void addEquivalent(T current) {
+		String uri = ((Resource) current).getURI();
+		if (uri != null) {
+			this.equivalents.add(uri);
+			setName(compareNames(uri, name));
+
+			if (current instanceof OntProperty) {
+				List<? extends OntResource> domainList = ((OntProperty) current).listDomain().toList();
```
MichaelRoeder: Defining something like
```java
OntProperty currentProperty = current.asProperty();
```
would make the following lines much easier (i.e., remove all the single casts)

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -189,6 +253,96 @@ public OntModel readOntology(String filePath) {
...
+		Set<Equivalent> elements = new HashSet<Equivalent>();
+
+		while (stack.size() > 0) {
+			T current = stack.pop();
+			boolean isSame = false;
+			boolean isVisited = visitedMap.get(current.toString());
+			if (!isVisited) {
+				List eqsList = null;
+				try {
+					if (current instanceof OntProperty)
```
MichaelRoeder: See comments from above :wink:

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -189,6 +253,96 @@ public OntModel readOntology(String filePath) {
...
+					if (eqsList != null) {
+						node.addEquivalentGroup((Set) eqsList.stream().collect(Collectors.toSet()));
+					}
+					elements.add(node);
+					uriNodeMap.put(current.toString(), node);
+				}
+				visitedMap.put(current.toString(), true);
+			}
+		}
+		return uriNodeMap;
```
MichaelRoeder, line 327: Does your algorithm execute an inferencer which makes all the equivalence relationships explicit?

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -189,6 +253,96 @@ public OntModel readOntology(String filePath) {
...
+	@SuppressWarnings({ "rawtypes", "unchecked" })
+	public <T extends OntResource> Map<String, Equivalent> searchEquivalents(Set ontElements) {
+
+		Map<String, Equivalent> uriNodeMap = new HashMap<String, Equivalent>();
+
+		Stack<T> stack = new Stack<T>();
+		stack.addAll(ontElements);
+
+		Map<String, Boolean> visitedMap = (Map<String, Boolean>) ontElements.stream()
+				.collect(Collectors.toMap(Object::toString, visited -> false));
```
MichaelRoeder: I don't think that this map is really necessary. Wouldn't it sufficient to check `!uriNodeMap.contains(current.toString())`?

MichaelRoeder (changes requested):
The implementation looks pretty good. I would recommend some smaller changes.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
aalexandrasilva: Hopefully, yes! The idea was to treat all the properties/classes present in the ontology before the inference. This way, when the actual inference takes place in the RDF Model, all the domains and ranges of the equivalence relationships established, would be explicit in these "Equivalent" objects.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
aalexandrasilva: yes, you are right! I've changed it :)

PR #10, opened 2019-10-07
Linked Geo

aalexandrasilva:
- Added Linked Geo Dataset


src/main/java/org/aksw/simba/lemming/creation/LinkedGeoDataset.java
```diff
@@ -0,0 +1,76 @@
...
+					ColouredGraph graph = creator.processModel(geoModel);
+					if (graph != null) {
+						LOGGER.info("Generated graph of " + geoModel.size() + " triples");
+						graphs.add(graph);
+					}
+				}
+			}
+		} else {
+			LOGGER.error("Find no files in \"" + folder.getAbsolutePath() + "\". Aborting.");
+			System.exit(1);
```
MichaelRoeder, line 74: This is very harsh. It is fine in our application here. However, in other applications, you would try to throw an exception or something similar :wink:

src/main/java/org/aksw/simba/lemming/tools/GraphGenerationTest.java
```diff
@@ -114,10 +117,14 @@ public static void main(String[] args) {
...
-        }else{
+        }else if(dataset.equalsIgnoreCase("swdf")){
         	LOGGER.info("Loading SemanticWebDogFood...");
         	mDatasetManager = new SemanticWebDogFoodDataset();
         	datasetPath = SEMANTIC_DOG_FOOD_DATA_FOLDER_PATH;
+        } else {
+        	LOGGER.info("Loading LinkedGeo...");
+        	mDatasetManager = new LinkedGeoDataset();
+        	datasetPath = LINKED_GEO_DATASET_FOLDER_PATH;
         }
```
MichaelRoeder, line 130: shoudln't it be something like 
```java
        }else if(dataset.equalsIgnoreCase("lgeo")){
        	LOGGER.info("Loading LinkedGeo...");
        	mDatasetManager = new LinkedGeoDataset();
        	datasetPath = LINKED_GEO_DATASET_FOLDER_PATH;
        } else {
        	LOGGER.error("Got an unknown dataset name: \"{}\". Aborting", dataset);
        	return;
        }

File: src/main/java/org/aksw/simba/lemming/tools/PrecomputingValues.java [hunk 3]
```diff
@@ -61,9 +64,15 @@ public static void main(String[] args) {
...
+            USE_LINKED_GEO = false;
+        } else  if (args[0].equals("swdf")){
             USE_SEMANTIC_DOG_FOOD = true;
             USE_PERSON_GRAPH = false;
+            USE_LINKED_GEO = false;
+        } else {
+        	USE_SEMANTIC_DOG_FOOD = false;
+            USE_PERSON_GRAPH = false;
+            USE_LINKED_GEO = true;
         }
```
MichaelRoeder, line 79: The same as in `GraphGenerationTest.java`. I would suggest a "defensive" programming and check whether linked geo data was given as argument or whether it is an unknown or mistyped dataset name.

MichaelRoeder (changes requested):
just some small comments that should be fixed. Apart from that, it already looks good.

src/main/java/org/aksw/simba/lemming/creation/LinkedGeoDataset.java
aalexandrasilva: should I also change the same in the other datasets then? :)

MichaelRoeder: No, as I wrote above, it is fine for this application. It was only a hint :wink:

aalexandrasilva:
Closed since the changes are present in the latest pull request: https://github.com/dice-group/Lemming/pull/16

PR #20, opened 2021-05-23
Improve precomputing values

wangzun66:
- [x] Inferer


src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -198,15 +217,15 @@ public void iterateStmts(Model newModel, Model sourceModel, OntModel ontModel, M
 			List<? extends OntResource> domain = property.listDomain().toList();
 			for (OntResource curResource : domain) {
 				Statement subjType = ResourceFactory.createStatement(subject, RDF.type, curResource);
-				if (!newModel.contains(subjType)) {
+				if (!model.contains(subjType)) {
```
aalexandrasilva: I would say these `model.contains(stmt);` checks can also be removed, since Model is a set of statements: [Jena Statements](https://jena.apache.org/tutorials/rdf_api.html)

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -153,9 +172,9 @@ private void checkEmptyTypes(Set<Resource> set, Model model) {
 	 * 
 	 * @param newModel    model where we will add the new triples
 	 * @param sourceModel provided model where we iterate through the statements
-	 * @param ontModel    the ontology model
+	 * @param uriNodeMap  the ontology model
```
aalexandrasilva: The `uriNodeMap` Javadoc is outdated.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -198,15 +217,15 @@ public void iterateStmts(Model newModel, Model sourceModel, OntModel ontModel, M
...
+				if (!model.contains(subjType)) {
 					newStmts.add(subjType);
 				}
 			}
 			if (object.isResource()) {
 				List<? extends OntResource> range = property.listRange().toList();
 				for (OntResource curResource : range) {
 					Statement objType = ResourceFactory.createStatement(object.asResource(), RDF.type, curResource);
-					if (!newModel.contains(objType)) {
+					if (!model.contains(objType)) {
```
aalexandrasilva: Same thing as the comment regarding `model.contains(stmt);`

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -239,27 +258,27 @@ public void iterateStmts(Model newModel, Model sourceModel, OntModel ontModel, M
...
+				ModelUtil.replaceStatement(model, statement,
 						ResourceFactory.createStatement(subject, newPredicate, object));
 		}
 
 		if (property != null) {
 			List<? extends OntResource> domain = property.listDomain().toList();
 			for (OntResource curResource : domain) {
 				Statement subjType = ResourceFactory.createStatement(subject, RDF.type, curResource);
-				if (!newModel.contains(subjType) && !curResource.isAnon()) {
+				if (!model.contains(subjType) && !curResource.isAnon()) {
```
aalexandrasilva: Same thing as the comment regarding `model.contains(stmt);`

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -239,27 +258,27 @@ public void iterateStmts(Model newModel, Model sourceModel, OntModel ontModel, M
...
 					newStmts.add(subjType);
 				}
 			}
 			if (object.isResource()) {
 				List<? extends OntResource> range = property.listRange().toList();
 				for (OntResource curResource : range) {
 					Statement objType = ResourceFactory.createStatement(object.asResource(), RDF.type, curResource);
-					if (!newModel.contains(objType)) {
+					if (!model.contains(objType)) {
```
aalexandrasilva: Same thing as the comment regarding `model.contains(stmt);`

aalexandrasilva:
@wangzun66 You might want to merge the develop branch into your working branch. It has a recent fix for an `IndexOutOfBounds` exception found while generating mimic graphs for the SWDF dataset.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
wangzun66: But <code>subjType</code> is also a statement, maybe we need another name without confusion.

wangzun66: But subjType is also a statement, maybe we need another name without confusion, like subjNewStmt, objNewStmt

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
wangzun66: Done

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
aalexandrasilva: What I meant to say is that since Jena's model is a Set of statements, duplicate statements have no effect on it. Thus, you don't need to check whether the model contains the statement already or not. 

> An RDF Model is represented as a set of statements. (Because a Model is set of statements, adding a duplicate of a statement has no effect.) 
From [Jena Model Tutorial](https://jena.apache.org/tutorials/rdf_api.html#ch-Statements)

wangzun66 (commented):
> @wangzun66 You might want to merge the develop branch into your working branch. It has a recent fix for an `IndexOutOfBounds` exception found while generating mimic graphs for the SWDF dataset.

@aalexandrasilva Which branch has IndexOutOfBounds exception？pgmerge or improve_PrecomputingValues

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
wangzun66: Ah, got it

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
wangzun66: Done

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
wangzun66: Done

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
wangzun66: Done

aalexandrasilva:
@wangzun66 The bug was found in the master branch, so it's likely both branches suffer from it.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -301,67 +316,67 @@ public OntModel readOntology(String filePath, String base) {
...
+					if(uriNodeMap.containsKey(curURI)){
+						curNode = uriNodeMap.get(curURI);
+					}
+					Map<T, Equivalent> localMap = new HashMap<>();
+					for(T re : eqsList){
+						if(re.getURI()!=null && uriNodeMap.containsKey(re.getURI())){
+							localMap.put(re, uriNodeMap.get(re.getURI()));
+						}
+					}
+					if(!localMap.isEmpty()){
```
aalexandrasilva: It seems redundant to me to check if the map is empty before iterating.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -301,67 +316,67 @@ public OntModel readOntology(String filePath, String base) {
...
+								curNode = localMap.get(re);
+							}else if (curNode != localMap.get(re)){
+								curNode.addEquivalentGroup(localMap.get(re).getEquiResources());
+							}
+						}
+					}
+					if(curNode==null){
+						curNode = new Equivalent(currentResource);
+					}
+					curNode.addEquivalent(currentResource);
```
aalexandrasilva: In the case that `curNode` was just created from `currentResource`, then trying to add `currentResource` again seems unnecessary. Since you already have the if statement in place right above, why not move `curNode.addEquivalent(currentResource);` to its else condition?

src/main/java/org/aksw/simba/lemming/creation/Equivalent.java
```diff
@@ -31,11 +31,17 @@
 	 */
 	private Set<String> equivalents = new HashSet<>();
 
+	/**
+	 * all equivalent-resources
+	 */
+	private Set<T> equiResources = new HashSet<>();
```
aalexandrasilva: Why do we need to keep track of both the resource objects and their respective strings? Wouldn't it be better to keep track of one or the other?

src/main/java/org/aksw/simba/lemming/creation/Equivalent.java
```diff
@@ -49,7 +55,9 @@ public void addEquivalent(T current) {
 		String uri = current.getURI();
 		
 		if (uri != null) {
-			this.equivalents.add(uri);
+			if(this.equivalents.add(uri)){
+				this.equiResources.add(current);
+			}
```
aalexandrasilva: You can add `this.equiResources.add(current);` directly without it being inside an if condition.

src/main/java/org/aksw/simba/lemming/creation/GraphCreator.java
```diff
@@ -274,18 +277,18 @@ protected ColourPalette createEdgePalette(Model model) {
 				}
 				// add the hierarchy information
 				// if there is no list of child nodes
```
aalexandrasilva: Please update the comments according to the changes you made.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -301,67 +316,67 @@ public OntModel readOntology(String filePath, String base) {
...
+					curNode.addEquivalentGroup(eqsList.stream().collect(Collectors.toSet()));
+					for(Object s : curNode.getEquivalents()){
+						uriNodeMap.put((String) s, curNode);
 					}
 				}
 			}
 		}
 		return uriNodeMap;
 	}
```
aalexandrasilva: I wonder if this method cannot be further optimized. 
I think some OWL Jena Reasoner can infer all the missing equivalentProperty and equivalentClass statements from the ontology. See  [Jena OWL Inference](https://jena.apache.org/documentation/inference/#OWLcoverage). However, the number of vertices of each dataset's target graph is likely to change with a different reasoner and would need to be updated.

If we change the reasoner with which the ontology is read with from the default one to ` ModelFactory.createOntologyModel(OntModelSpec.OWL_LITE_MEM_RULES_INF);`  in the dataset managers. It would simplify this searchEquivalents method, since we would only need to:
```
for (T currentResource : ontElements) {
  //  list equivalent resources of currentResource 
  //  create an Equivalent object, add the resources obtained in the previous step and update uriNodeMap 
}
```
@MichaelRoeder what do you think?

MichaelRoeder: It would be nice if we could simply use an existing implementation. However, my experience with the Jena implementations are not good. They follow very strictly the theoretical rules and add a lot of information that we are not interested in. Hence, they will be slower and increase the size of our ontology a lot.

However, I agree that the method can be optimized further. Apart from smaller improvements (e.g., line 371 is a very costly transformation of a list to a set), I am also afraid that it could contain a severe bug. If I understand the code correctly (it is not so easy to understand the work flow of the method so I am not 100% sure :sweat_smile: ), line 373 could override information that was already in the map but will be lost since the `Equivalent` instances of all other URIs are overridden with the information that we get from the current URI. This does not take into account that there might have been information added before. :thinking: 

First, I would like to clarify whether we really need the class `Equivalent`. To me, it seems to make things more complicated and it also seems to consume more space than necessary :thinking: 
I saw that it also overrides domain and range information of properties. I think this isn't something that such a class should do. Instead, the Inferer should do that in a separate method (after it identified the properties that are equivalent to each other).

Wouldn't a `Map<T, Set<T>>` be sufficient to store equivalents?

aalexandrasilva: The `Equivalent` class was created to store the set of equivalent classes/properties, to store an "uniform" URI that would be used to replace the equivalent classes/properties in statements and to keep track of the domain and range information of all equivalent properties. 

But yes, if we pass these responsibilities to the Inferer, then it should also be enough to have a `Map<T, Set<T>>` to store the equivalents.

MichaelRoeder: I see. Well, in that case, I would suggest to remove the class and separate these steps, i.e.,
1. Search for equivalent IRIs with `Map<T, Set<T>>`
2. For each set of equivalent IRIs, choose an IRI that represents it. (for properties, ensure that domain and range are adapted)

The second step means that classes have to be handled first right? Otherwise we end up with classes in our domain and range definition, that may have been replaced with another class :thinking:

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -198,70 +216,89 @@ public void iterateStmts(Model newModel, Model sourceModel, OntModel ontModel, M
...
+					}
+				}
+			}
+			//find a name to represent all equivalent properties
+			String name = null;
+			for(OntProperty ontPro : eqsProperties){
+				if(name == null || name.compareTo(ontPro.getURI())<0){
+					name = ontPro.getURI();
+				}
+			}
```
aalexandrasilva: You just finished iterating through `eqsProperties`, so you don't need to iterate again to search for the name.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -198,70 +216,89 @@ public void iterateStmts(Model newModel, Model sourceModel, OntModel ontModel, M
...
+			Property newPredicate = ResourceFactory.createProperty(name);
 			if (!newPredicate.getURI().equals(predicate.getURI()))
-				ModelUtil.replaceStatement(newModel, statement,
+				ModelUtil.replaceStatement(model, statement,
 						ResourceFactory.createStatement(subject, newPredicate, object));
 		}
```
aalexandrasilva: I don't think the `searchType` method is the best place to gather the domain and range of the equivalents, you would end up repeating this operation unnecessarily, since `searchType` is performed inside a loop of all the model's statements.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -290,90 +327,93 @@ public OntModel readOntology(String filePath, String base) {
...
+					}
+					for(T eq : eqsList){
+						if(eq.getURI() != null){
+							equiSet.add(eq);
+						}
+					}
+					for(T s : equiSet){
+						iriEquiMap.put(s, equiSet);
 					}
 				}
```
aalexandrasilva: 1) It looks like you don't need the `localMap` at all, since you're just adding the same information as in `iriEquiMap` (L369 `localMap.put(re, iriEquiMap.get(re));`)? 
2) You have 3 loops going through `eqsList` in succession, are these really required?
3) Since all equivalent properties will point to the same `Set<T>`, I think you would only need to update one of the references to this Set, there wouldn't be a need to update every map entry.

Please tell me if I misunderstood something :smiley:

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -198,70 +216,89 @@ public void iterateStmts(Model newModel, Model sourceModel, OntModel ontModel, M
...
 
-		Equivalent<OntProperty> node = uriNodeMap.get(predicate.toString());
+		Set<OntProperty> eqsSet = iriEquiMap.get(predicate);
 		OntProperty property = null;
+		if(eqsSet != null){
+			List<OntProperty> eqsProperties = new ArrayList<>(iriEquiMap.get(predicate));
+			property = eqsProperties.get(0);
+			//Add all domains and ranges of equivalent properties to property
+			for(int i=1; i<eqsProperties.size(); i++){
+				OntProperty currentProperty = eqsProperties.get(i);
```
aalexandrasilva: According to the `searchEquivalents` method, it is possible for the `iriEquiMap.get(predicate)` to be empty and your code will throw an exception in this case. 
From what I understand, I don't think you need to create a new list, you could just iterate the set directly and it would be safe to use if it's empty: 

```
Set<OntProperty> eqsProperties = iriEquiMap.get(predicate);
for(OntProperty currentProperty : eqsProperties){
	if(property == null) {
		property = currentProperty;
		continue;
	}
        ...
}
```

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -41,67 +32,93 @@
...
+			//if the model doesn't grow, terminate the loop
+			}while(curSize != newSize);
 		}
+
+		// infer type statements, a single property name is also enforced here
+		iterateStmts(newModel, sourceModel, this.propertiesEquiMap);
+		checkEmptyTypes(set, newModel);
+
+		// uniform the names of the classes
+		renameClasses(newModel, this.classesEquiMap);
```
aalexandrasilva: As Michael pointed out, we should move the class renaming before the inference.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
wangzun66: The localMap is for the transitive case: if A <=>B <=> C and A, C are visited firstly, then we would get two equiSet (A, B) and (C, B). The function of localMap is to collect such equiSets, and merge them to an object, and update map entries A, B, C using the merged equiSet.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
wangzun66: Sure, it is indeed a very bad place. I've modified that to ensure that domain and range are calculated only one time.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
aalexandrasilva: Hm from what I see `eqsList` is essentially your set of keys for this `localMap`. 
Then why do you need to have `localMap` if you just add the exact same information present in `iriEquiMap` for this eqsList keyset? Why not work directly with `iriEquiMap`?

MichaelRoeder (changes requested):
There are still comments that have not been addressed.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
wangzun66: Done

src/main/java/org/aksw/simba/lemming/creation/GraphCreator.java
wangzun66: Done

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
wangzun66: Done

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
wangzun66: I've changed the code, there's no such problem now.

wangzun66:
> There are still comments that have not been addressed.
@MichaelRoeder Now all comments have been addressed.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -290,93 +309,131 @@ public OntModel readOntology(String filePath, String base) {
...
-				ResourceUtils.renameResource(mResource, newName);
+			Map.Entry<OntClass, String> pair = it.next();
+			Resource mResource = model.getResource(pair.getKey().getURI());
+			if (mResource != null && !mResource.getURI().equals(pair.getValue())) {
+				ResourceUtils.renameResource(mResource, pair.getValue());
+			}
+		}
+	}
+
+	private <T extends OntResource> Map<T, String> findProperName(Map<T, Set<T>> iriEquiMap){
```
aalexandrasilva: Missing documentation.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -290,93 +309,131 @@ public OntModel readOntology(String filePath, String base) {
...
+				}
+				for(T eq : iriEquiMap.get(re)){
+					iriToName.put(eq, properName);
+				}
+			}
+		}
+		return  iriToName;
+	}
+
+	private Map<OntProperty,OntProperty> collectRangeDomain(Map<OntProperty, Set<OntProperty>> propertyEquiMap){
```
aalexandrasilva: Missing documentation.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -290,93 +309,131 @@ public OntModel readOntology(String filePath, String base) {
...
+		Map<OntProperty,OntProperty> ontProToOntProDR = new HashMap<>();
+		for(OntProperty property : propertyEquiMap.keySet()){
+			List<OntProperty> eqsProperties = new ArrayList<>(propertyEquiMap.get(property));
+			OntProperty collectedProperty = eqsProperties.get(0);
+			for(OntProperty currentProperty : eqsProperties){
```
aalexandrasilva: Again, `OntProperty collectedProperty = eqsProperties.get(0);` might lead to an exception. Also you don't need to convert the set to a list. There's a mention to this same problem in one of the open comments.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
aalexandrasilva: I think this problem hasn't been solved, only the location of the code changed.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
wangzun66: Sure. But I still need an object OntProperty from the set to carry all domains and ranges. Maybe before creating a list, check first if the set is empty or not. If so, skip the whole iteration.

wangzun66: Set<OntProperty> equiSet = propertyEquiMap.get(property);
			if(equiSet.isEmpty()){
				continue;
			}
			List<OntProperty> eqsProperties = new ArrayList<>(equiSet);
			OntProperty collectedProperty = eqsProperties.get(0);

wangzun66: I've added all domains and ranges on the key itself from the given map.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -290,93 +309,140 @@ public OntModel readOntology(String filePath, String base) {
...
+				List<? extends OntResource> rangeList = currentProperty.listRange().toList();
+				for (OntResource range : rangeList) {
+					if (!collectedProperty.hasRange(range)) {
+						collectedProperty.addRange(range);
+					}
+				}
 			}
+			ontProToOntProDR.put(property, collectedProperty);
 		}
+		return ontProToOntProDR;
```
aalexandrasilva: Given three equivalent properties A, B and C. I am assuming the `propertyEquiMap` contents would look something like:
```
A -> {A, B, C}
B -> {A, B, C}
C -> {A, B, C}
```
So I think you might be collecting the domain and range of properties which could have already been collected. 
One suggestion would be to just use the properties with the _proper name_ (the output of the `findProperName` method) to collect the domain and range information on the corresponding object of this uniform URI. That way you would only need to collect this information once per set of equivalents instead of for all properties.

aalexandrasilva (commented):
The project is not building correctly at the moment, as there are compilation errors in `PrecomputingValuesTest`. 

I think a unit test case for the equivalent resources search would be nice to have. @MichaelRoeder What do you think?

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
wangzun66: > The project is not building correctly at the moment, as there are compilation errors in `PrecomputingValuesTest`.
> 
> I think a unit test case for the equivalent resources search would be nice to have. @MichaelRoeder What do you think?

@aalexandrasilva Yes, the PrecomputingValuesTest is failed, because it only tests whether the generated models with the new Inferer are the same as the old inferer. Now they are not same, because the order of iterateStmts and renameClasses in Inferer are exchanged. I think, maybe the PrecomputingValueTest could be removed.

aalexandrasilva: From what I see it comes from a compilation error because `Equivalent` class has been deleted.

aalexandrasilva: From what I see the error comes from a compilation error since the Equivalent class has been deleted.
However, I do agree with removing PrecomputingValueTest and have more unit test cases instead to check the functionality. But we should wait on @MichaelRoeder's opinion on that.

MichaelRoeder: I fully agree with @aalexandrasilva. It does not make much sense to keep the PrecomputingValueTest if it solely relies on the comparison of the old and the new Inferer implementations. It makes much more sense to implement tests that focus on the functionality that we want to achieve.

MichaelRoeder:
@wangzun66 from the discussion above, I gathered that this PR still needs to be fixed. :wink:

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
```diff
@@ -274,13 +249,13 @@ public void iterateStmts(Model newModel, Model sourceModel, OntModel ontModel, M
 	 * @param filePath path to the ontology file
 	 * @return OntModel Object
 	 */
-	public OntModel readOntology(String filePath, String base) {
+	private OntModel readOntology(String filePath, String base) {
```
MichaelRoeder: The term "base" is misleading here. With "base", the jena framework typically means the base IRI of a file. However, it looks like it is used here as a language identifier. Please rename the variable (and add it to the javadoc comment).

MichaelRoeder (commented):
LGTM.
@wangzun66  I only have added one comment that should be fixed.

@aalexandrasilva can you please have a final look at the PR?

aalexandrasilva:
It also looks good to me.

aalexandrasilva (commented):
It also looks good to me.

src/main/java/org/aksw/simba/lemming/creation/Inferer.java
wangzun66: I changed the name to "fileType".

MichaelRoeder (approved):
LGTM

PR #22, opened 2021-05-30
Optimize maxvertexdegree

atulpundir88:
Change for improving the run time of MaxVertexDegreeMetric.
[Document.docx](https://github.com/dice-group/Lemming/files/6566978/Document.docx)


src/main/java/org/aksw/simba/lemming/metrics/single/edgemanipulation/EdgeModifier.java
```diff
@@ -86,6 +89,8 @@ public ColouredGraph getGraph(){
 			
 			ObjectDoubleOpenHashMap<String> mapChangedMetricValues = new ObjectDoubleOpenHashMap<String>();
 			
+			metricComputations.removeEdgeFromGraph(triple);//remove edge in metric computation graph
+
```
MichaelRoeder: The `metricComputations` and the `mEdgeModification` object share the same graph object. If one of the removes an edge, it is also removed for the other object. I don't see why this triple has to be removed twice.

src/main/java/org/aksw/simba/lemming/metrics/single/edgemanipulation/EdgeModifier.java
```diff
@@ -116,6 +122,8 @@ public ColouredGraph getGraph(){
 	        		triple.tailColour, removedEdgeID, triple.edgeColour);
 	        mLstRemovedEdges.add(removedTriple);
 	        
+	        metricComputations.addEdgeToGraph(triple);//add edge in metric computation graph
```
MichaelRoeder: See comment above.

src/main/java/org/aksw/simba/lemming/metrics/single/edgemanipulation/EdgeModifier.java
```diff
@@ -131,6 +139,8 @@ public ColouredGraph getGraph(){
 			ObjectDoubleOpenHashMap<String> mapMetricValues = new ObjectDoubleOpenHashMap<String>();
 			triple.edgeId = mEdgeModification.addEdgeToGraph(triple.tailId,triple.headId, triple.edgeColour);
 			
+			metricComputations.addEdgeToGraph(triple);//add edge in metric computation graph
+
```
MichaelRoeder: See comment above.

src/main/java/org/aksw/simba/lemming/metrics/single/edgemanipulation/MetricComputations.java
```diff
@@ -0,0 +1,321 @@
...
+				graphOperation = graphOperation + "Outdegree";
+				metVal = metricComputationMaxDegree(metric, graph, graphOperation, DIRECTION.out, triple.tailId, triple, checkIfRemoveOrAddEdge(graphOperation));
+				break;
+
+			default:// If metric is other than maxInDegree and maxOutDegree then apply the metric
+				metVal = metric.apply(graph);
+		}
+
+		return metVal;
+	}
```
MichaelRoeder: The usage of `graphOperation` in form of a string is costly. Please think of a better way to distinguish the cases (e.g., a flag or an enum—both approaches that are much easier to handle by a machine).

src/main/java/org/aksw/simba/lemming/metrics/single/edgemanipulation/EdgeModifier.java
```diff
@@ -146,14 +156,17 @@ public ColouredGraph getGraph(){
...
+        			//double metVal = metric.apply(graph);
+        			double metVal = metricComputations.getMetricValue(triple, metric, graph, "AddAnEdge");
 	        		mapMetricValues.put(metric.getName(), metVal);   
 	        	}
 	        }
 		    
 		    //mEdgeModification.removeEdgeFromGraph(triple.edgeId);
 		    mEdgeModification.removeEdgeFromGraph(triple.edgeId, (int)mMapMetricValues.get("#nodetriangles"), 
 		    		(int)mMapMetricValues.get("#edgetriangles"));
+		    metricComputations.removeEdgeFromGraph(triple);//remove edge in metric computation graph
```
MichaelRoeder: See comment above.

src/main/java/org/aksw/simba/lemming/metrics/single/edgemanipulation/EdgeModifier.java
```diff
@@ -178,6 +191,10 @@ public void executeRemovingAnEdge(ObjectDoubleOpenHashMap<String> newMetricValue
 			mEdgeModification.removeEdgeFromGraph(lastTriple.edgeId, (int) newMetricValues.get("#nodetriangles"),
 					(int) newMetricValues.get("#edgetriangles"));
 			
+			metricComputations.setCandidatesMetricRemoveAnEdge( metricComputations.getmMapCandidatesMetricTemp(), metricComputations.getmMapCandidatesMetricValuesTemp());
+			metricComputations.removeEdgeFromGraph(lastTriple);
```
MichaelRoeder: See comment above.

src/main/java/org/aksw/simba/lemming/metrics/single/edgemanipulation/EdgeModifier.java
```diff
@@ -197,6 +214,10 @@ public void executeAddingAnEdge(ObjectDoubleOpenHashMap<String> newMetricValues)
 			mEdgeModification.addEdgeToGraph(lastTriple.tailId, lastTriple.headId, 
 									lastTriple.edgeColour, (int) newMetricValues.get("#nodetriangles"),
 									(int) newMetricValues.get("#edgetriangles"));
+			
+			metricComputations.setCandidatesMetricAddAnEdge( metricComputations.getmMapCandidatesMetricTemp(), metricComputations.getmMapCandidatesMetricValuesTemp());
+			metricComputations.addEdgeToGraph(lastTriple);
```
MichaelRoeder: See comment above.

MichaelRoeder (changes requested):
As already discussed, the changes go into the right direction. However, I would highly suggest to keep the details of the metrics within the metrics classes instead of having an additional class that has to take care of the different metric types.

I will describe a possible way in the following. Please note that this is a draft. You would have to add the implementation to the single metric classes (including javadocs that I have skipped here!).

The main idea is that the metrics class is enabled to built upon its previous result. This could be done by extending the `SingleValueMetric` interface with the following two methods:
```java
    public default UpdatableMetricResult applyUpdatable(ColouredGraph graph) {
        return new SimpleMetricResult(getName(), apply(graph));
    }
    
    public default UpdatableMetricResult update(TripleBaseSingleID triple, SingleValueMetric metric, ColouredGraph graph,
            boolean graphOperation, UpdatableMetricResult previousResult) {
        return applyUpdatable(graph);
    }
```
The first method allows the metric implementation to return an object that may contain information that is necessary for a later update (in addition to its metric result). The second method implements the update method (note that the `graphOperation` argument is a boolean instead of a String. As described in the comments, an enum could work as well). Both methods offer a `default` implementation which ensures backwards compatibility.

The `UpdatableMetricResult` is an interface defined as follows:
```java
public interface UpdatableMetricResult {

    public double getResult();
    public String getMetricName();
}
```
It should be noted that this interface does not really define something special. However, it allows the implementation of metric specific classes that can store additional information, e.g., the ID of the node that has the highest degree.

To ensure the backward compatibility, a basic implementation of this interface is defined as follows:
```java
public class SimpleMetricResult implements UpdatableMetricResult {

    protected String metricName;
    protected double result;
    
    public SimpleMetricResult(String metricName, double result) {
        this.metricName = metricName;
        this.result = result;
    }

    @Override
    public double getResult() {
        return result;
    }

    @Override
    public String getMetricName() {
        return metricName;
    }

    public void setMetricName(String metricName) {
        this.metricName = metricName;
    }

    public void setResult(double result) {
        this.result = result;
    }
}
```
Based on these interfaces, the implementation of the amendment phase can keep `UpdatableMetricResult` objects for each metric. These kept objects can be used to call the `update` method of the single metrics instead of their `apply` method. Depending on the metric's implementation, the update method might be able to reuse the previous results instead of starting from scratch.

atulpundir88:
I have included the changes suggested by the above comments. Also, committed the changes which optimize the execution of the Average vertex degree metric. 
[Document.docx](https://github.com/dice-group/Lemming/files/6643596/Document.docx)

MichaelRoeder:
Please ensure to stick to the formatting of the project. To this end, I would suggest to copy the `.editorconfig` file from https://github.com/dice-group/orca/blob/master/.editorconfig and to configure your IDE to work with it.

src/main/java/org/aksw/simba/lemming/metrics/single/SingleValueMetric.java
```diff
@@ -1,22 +1,61 @@
...
+     * 
+     * @param graph
+     *            the graph for which the metric should be calculated.
+     * @return the value of the metric.
+     */
+    public double apply(ColouredGraph graph);
+    
+    
+	/**
+	 * Returns metric results that for further computations could be reused.
```
MichaelRoeder: The comment is grammatically wrong. It should be `Returns metric results that can be reused for further computations.`

src/main/java/org/aksw/simba/lemming/metrics/single/SingleValueMetric.java
```diff
@@ -1,22 +1,61 @@
...
+	 * 
+	 * @param graph
+	 *            - input graph.
+	 * @param prevMetricResult
+	 *            - UpdatableMetricResult object containing the results that could
+	 *            be used.
+	 * @return - metric result.
+	 */
+	public default UpdatableMetricResult applyUpdatable(ColouredGraph graph, UpdatableMetricResult prevMetricResult) {
```
MichaelRoeder: The signature of this method is still wrong. It means that for creating an instance of an `UpdatableMetricResult` another `UpdatableMetricResult` is needed. However, that is not the case. This method should simply be the `apply` method with a different return type. You can also see that from the default implementation.

src/main/java/org/aksw/simba/lemming/metrics/single/MaxVertexDegreeMetric.java
```diff
@@ -13,20 +19,261 @@
...
+						triple.tailId, triple, 1, previousResult, mVertexDegrees);
+			} else {
+				newMetricResult = metricComputationMaxDegree(graph, GRAPHOPERATION.RemoveAnEdgeOutdegree, DIRECTION.out,
+						triple.tailId, triple, -1, previousResult, mVertexDegrees);
+			}
+			break;
+
+		default:// If metric is other than maxInDegree and maxOutDegree then apply the metric
+			newMetricResult = applyUpdatable(graph, previousResult);
+		}
```
MichaelRoeder: I do not really understand why you rely on the two names. You should rely on the direction as the apply method shows.

src/main/java/org/aksw/simba/lemming/metrics/single/AvgVertexDegreeMetric.java
```diff
@@ -1,37 +1,151 @@
...
+	 */
+	@Override
+	public UpdatableMetricResult update(TripleBaseSingleID triple, ColouredGraph graph, boolean graphOperation,
+			UpdatableMetricResult previousResult, VertexDegrees mVertexDegrees) {
+
+		if (getName().equals("stdDevInDegree") || getName().equals("stdDevOutDegree")) {
+			return new SimpleMetricResult(getName(), apply(graph));
+		}
+
+		AvgVertexDegreeMetricResult metricResultTempObj = (AvgVertexDegreeMetricResult) applyUpdatable(graph, previousResult);
```
MichaelRoeder: There is a general misunderstanding of the meaning of the two newly added methods. Calling applyUpdatable within the update method does not really make sense to me. Please check again the description of the methods.

src/main/java/org/aksw/simba/lemming/metrics/single/AvgVertexDegreeMetric.java
```diff
@@ -1,37 +1,151 @@
...
+			}
+
+			metricResultTempObj.setSumVertexDeg(sum);
+			metricResultTempObj.setNumberOfVertices(numberOfVertices);
+		}
+
+		sum = sum / numberOfVertices; // Compute Metric value
+		metricResultTempObj.setResult(sum);
+
+		return metricResultTempObj;
```
MichaelRoeder: This can be further optimized in an easy way. Please think about which piece of information you _really_ need when you want to get the average degree, degree standard deviation or one of the minimum or maximum values.

src/main/java/org/aksw/simba/lemming/metrics/single/edgemanipulation/EdgeModifier.java
```diff
@@ -27,188 +29,218 @@
...
+
+			ColouredGraph graph = mEdgeModification.getGraph();
+
+			for (SingleValueMetric metric : mLstMetrics) {
+				if (!metric.getName().equalsIgnoreCase("#edgetriangles")
+						&& !metric.getName().equalsIgnoreCase("#nodetriangles")) {
+					// double metVal = metric.apply(graph);
+					// Calling update method to get the metric values based on previous results
+					mMapPrevMetricsResultRemoveEdge.put(metric.getName(), metric.update(triple, graph, false,
+							new HashMap<>(mMapPrevMetricsResult).get(metric.getName()), mEdgeModification.getmVertexDegrees()));
```
MichaelRoeder: Why is there a new hash map created? This is very costly and not necessary!

src/main/java/org/aksw/simba/lemming/metrics/single/edgemanipulation/VertexDegrees.java
```diff
@@ -0,0 +1,181 @@
...
+/**
+ * Class for storing in and out degree of vertices.
+ * 
+ * @author Atul
+ *
+ */
+public class VertexDegrees {
+
+	// Maps for storing in-degree and out-degree for different vertices
+	private int mMapVerticesinDegree[];
```
MichaelRoeder: Please make sure that you use camelCase writing. In several attribute and method names, not all words within the name start with a capital letter. In this line for example, the `in` should be `In`.

MichaelRoeder (changes requested):
Several files still have the wrong formatting (e.g., they use tab characters for the intend instead of white spaces). Apart from that, there seems to be a misunderstanding of the methods described by me above.

src/main/java/org/aksw/simba/lemming/metrics/single/SingleValueMetric.java
atulpundir88: Done. I have corrected the comment.

src/main/java/org/aksw/simba/lemming/metrics/single/SingleValueMetric.java
atulpundir88: Done. I have removed the UpdatableMetricResult parameter.

src/main/java/org/aksw/simba/lemming/metrics/single/MaxVertexDegreeMetric.java
atulpundir88: Done. I have changed it, now using direction instead of metric names.

src/main/java/org/aksw/simba/lemming/metrics/single/AvgVertexDegreeMetric.java
atulpundir88: okay. I have removed the call to applyUpdatable method, using the previousResult object to get the old computed values.

src/main/java/org/aksw/simba/lemming/metrics/single/edgemanipulation/EdgeModifier.java
atulpundir88: I have removed this logic and using the values stored in the mMapPrevMetricsResult map.

src/main/java/org/aksw/simba/lemming/metrics/single/edgemanipulation/VertexDegrees.java
atulpundir88: Done. I have corrected the variable and method names.

src/main/java/org/aksw/simba/lemming/metrics/single/AvgVertexDegreeMetric.java
atulpundir88: For the standard deviation metric, logic was created in StdDevVertexDegree class (update method) to calculate it in an optimized way. These changes are implemented by Pranav and are available in the optimize_stddev_metric branch. Thus, this method will not be called for standard deviation metric once those changes are included. We are checking further to implement it in a better way.

MichaelRoeder (changes requested):
The formatting of the code files is still different than the original formatting. This introduces a lot of problems. Please make sure that you use 4 whitespaces for the indentation instead of a tab character. You can configure it in your IDE. Please do so, format all the files changed in this PR and commit them.

src/main/java/org/aksw/simba/lemming/metrics/single/AvgVertexDegreeMetric.java
```diff
@@ -1,37 +1,128 @@
...
+
+            if (metricResultTempObj.getSumVertexDeg() == 0.0) {
+                // Computing the Avg Vertex Degree Metric for the first time
+
+                int[] getmMapVerticesinDegree = mVertexDegrees.getMapVerticesInDegree();
+
+                for (int key = 0; key < getmMapVerticesinDegree.length; key++) { // Compute sum in iteration
+                    sum += getmMapVerticesinDegree[key];
+                }
+                numberOfVertices = getmMapVerticesinDegree.length;
```
MichaelRoeder: This is exactly the same as it is done in lines 77-85, right? And it is also the same functionality, that is done when this metric is called without updating a given metric object. Typically, it is much better to have exactly 1 piece of code that measures it and call it from all places at which it is needed.

src/main/java/org/aksw/simba/lemming/metrics/single/AvgVertexDegreeMetric.java
```diff
@@ -1,37 +1,128 @@
...
+    public UpdatableMetricResult update(TripleBaseSingleID triple, ColouredGraph graph, boolean graphOperation,
+            UpdatableMetricResult previousResult, VertexDegrees mVertexDegrees) {
+
+        AvgVertexDegreeMetricResult metricResultTempObj = new AvgVertexDegreeMetricResult(getName(), 0.0);
+        if (previousResult instanceof AvgVertexDegreeMetricResult) {
+            // Copying previously computed values in temporary variables
+            metricResultTempObj.setSumVertexDeg(((AvgVertexDegreeMetricResult) previousResult).getSumVertexDeg());
+            metricResultTempObj
+                    .setNumberOfVertices(((AvgVertexDegreeMetricResult) previousResult).getNumberOfVertices());
+        }
```
MichaelRoeder: In an else block, we could already calculate the value from scratch and leave the method. It is not necessary to go on because there is nothing we can update, right?

src/main/java/org/aksw/simba/lemming/metrics/single/AvgVertexDegreeMetricResult.java
```diff
@@ -0,0 +1,66 @@
...
+ * iterations.
+ * 
+ * @author Atul
+ *
+ */
+public class AvgVertexDegreeMetricResult extends SimpleMetricResult {
+
+    private double sumVertexDeg = 0.0; // Variable for storing sum of vertex degrees
+
+    private double numberOfVertices = 0.0; // Variable for storing number of vertices.
```
MichaelRoeder: The two comments should be javadoc comments :wink:

src/main/java/org/aksw/simba/lemming/metrics/single/MaxVertexDegreeMetric.java
```diff
@@ -29,4 +34,181 @@ public double apply(ColouredGraph graph) {
...
+     *            modified.
+     * @param direction
+     *            - this is in or out based on the operation.
+     * @param vertexID
+     *            - The vertex that is modified.
+     * @return
+     */
+    private UpdatableMetricResult metricComputationMaxDegree(ColouredGraph graph, 
+            DIRECTION direction, int vertexID, TripleBaseSingleID triple, int updateVertexDegree,
+            UpdatableMetricResult previousResult, VertexDegrees mVertexDegrees) {
```
MichaelRoeder: Please check the parameters of this method. It seems like `triple` is not used and `direction` is an attribute of this class.

src/main/java/org/aksw/simba/lemming/metrics/single/MaxVertexDegreeMetric.java
```diff
@@ -29,4 +34,181 @@ public double apply(ColouredGraph graph) {
...
+                    .setCandidatesMetricSet(((MaxVertexDegreeMetricResult) previousResult).getmMapCandidatesMetric());
+            metricResultTempObj.setMaxVertexDegree(
+                    ((MaxVertexDegreeMetricResult) previousResult).getMaxVertexDegree());
+        }
+
+        IntSet intSetTemp = metricResultTempObj.getmMapCandidatesMetric();
+        // Get the current candidate set
+
+        if (intSetTemp.size() == 0) { // Initially the Candidate set will be empty, hence need to call the apply
+                                      // method and store the candidates
```
MichaelRoeder: But why we are here if there is no initial result that can be updated? Shouldn't we have a different method for that? :confused:

MichaelRoeder (changes requested):
I started to review the changes. However, there seems to be a general misunderstanding how the metrics should be structured. The current implementations of metrics seem to have the following disadvantages:

1. Metrics classes have several places at which the metric value might be calculated. This has the disadvantage that you would have to write tests for all of them. It would be preferable that there is exactly one single method that calculates the value.
    1. Metrics still have their simple `apply` method implementation that just calculates the metric and returns it. That is not necessary.
    2. The `applyUpdatable` method does not seem to be used anywhere within the metric. That is quite confusing.
    3. The `update` method checks whether a given result can be updated. Otherwise, it will be calculated. Unfortunately, it seems like every `update` method implements its own calculation of the metric again...
2. It seems to be complicated for the metric to figure out whether the given `UpdatableMetricResult` object contains a result that can be used by this class. It seems like the metrics check for values that are `0` or have some other value they are looking for. From my point of view, a more common structure would be more helpful, e.g., if there is an object, it can be used. If the given variable is `null`, there is no previous result that could be reused.

It should be noted that we override the default implementations that have been defined in the `SingleValueMetric` class. That is fine because the default implementations are only there to make sure that a metric that can not be updated still can be used. However, it does not mean that the structure in the `SingleValueMetric` is something that is really helpful for an updateable structure. For such a metric, the general structure should be something like the following:
```java
    public double apply(ColouredGraph graph) {
        return applyUpdatable(graph).getResult();
    }

    public UpdatableMetricResult applyUpdatable(ColouredGraph graph) {
        // Calculate the metric result from scratch but return it 
        // as UpdatableMetricResult object that contains additional 
        // information
        return ...;
    }

    public UpdatableMetricResult update(TripleBaseSingleID triple, ColouredGraph graph, boolean graphOperation,
			UpdatableMetricResult previousResult, VertexDegrees mVertexDegrees) {
        // First, decide whether we have something that we can update.
        // As suggested, this could be a check for null
        if(previousResult == null) {
            // There is nothing to update, so we should calculate it again.
            // Let's use the method that we already have defined before!
            return applyUpdatable(graph);
        }

        // Figure out, whether we have a case in which we can simply update 
        // the given, previous result. If yes, update it and return ...
        // If not, again, use applyUpdatable(graph) to calculate the new result
    }
```

src/main/java/org/aksw/simba/lemming/metrics/single/AvgVertexDegreeMetric.java
atulpundir88: Done. Have merged code from Metrics_Optimization branch that was not included here and the logic is simplified now.

src/main/java/org/aksw/simba/lemming/metrics/single/AvgVertexDegreeMetric.java
atulpundir88: Done. Have merged code from Metrics_Optimization branch that was not included here and the logic is simplified now.

src/main/java/org/aksw/simba/lemming/metrics/single/AvgVertexDegreeMetricResult.java
atulpundir88: Done.

src/main/java/org/aksw/simba/lemming/metrics/single/MaxVertexDegreeMetric.java
atulpundir88: Done.

src/main/java/org/aksw/simba/lemming/metrics/single/MaxVertexDegreeMetric.java
atulpundir88: Done. This logic is removed now. Max vertex degree metric makes use of number of vertices having maximum degree and checks if the metric should be updated or not.

src/main/java/org/aksw/simba/lemming/metrics/single/SingleValueMetricResult.java
```diff
@@ -0,0 +1,68 @@
+package org.aksw.simba.lemming.metrics.single;
+
+/**
+ * The class stores the metric name and the result which can be reused.
+ * 
+ * @author Atul
+ *
+ */
+public class SingleValueMetricResult implements UpdatableMetricResult {
```
MichaelRoeder: Where is the difference between this class and the `SimpleMetricResult` class? If both classes are the same, please keep only one of them and remove the other.

MichaelRoeder (commented):
The changes look good so far. I think one of the classes can be removed (see the comment below).

src/main/java/org/aksw/simba/lemming/metrics/single/SingleValueMetricResult.java
atulpundir88: Yes, they are similar. I have removed SingleValueMetricResult class.

PR #23, opened 2021-06-11
Optimize stddev metric

Hollowfier:
Changes to StdDevVertexDegree metric to avoid calling its apply method each time

This branch also contains changes to  MaxVertexDegree made by Atul

------------------------------------------------------
Base case Runtime on Local machine:
Geology dataset: 61 mins

Optimized Runtime on Local machine:
Geology dataset: 27 mins

------------------------------------------------------
Base case Runtime on Virtual Environment:
Geology dataset: 41 mins
SWDF: 22 mins

Optimized Runtime on Virtual Environment:
Geology dataset: 35 mins
SWDF: 23 mins


src/main/java/org/aksw/simba/lemming/metrics/single/edgemanipulation/EdgeModification.java
```diff
@@ -258,6 +275,16 @@ public void removeEdgeFromGraph(int edgeId, int newNodeTriangles, int newEdgeTri
         this.newNodeTriangles = newNodeTriangles;
         this.newEdgeTriangles = newEdgeTriangles;
         this.graph.removeEdge(edgeId);
+        
+        //For removing an edge from a graph, update the vertex in and out degrees.
+        IntSet verticesConnectedToRemovingEdge = graph.getGraph().getVerticesIncidentToEdge(edgeId);
+        int headId = verticesConnectedToRemovingEdge.size() > 1 ? verticesConnectedToRemovingEdge.toIntArray()[1] 
+        		: verticesConnectedToRemovingEdge.toIntArray()[0];
+        int tailId = verticesConnectedToRemovingEdge.toIntArray()[0];
```
MichaelRoeder: wouldn't be something like this easier and faster?
```java
int tail = getDirectedSimpleEdgeTail(e);
int head = getDirectedSimpleEdgeHead(e);
```

src/main/java/org/aksw/simba/lemming/metrics/single/edgemanipulation/EdgeModifier.java
```diff
@@ -27,6 +33,8 @@
 	private List<TripleBaseSingleID> mLstAddedEdges;
 	private boolean isCoutingEdgeTriangles = false;
 	private boolean isCountingNodeTriangles = false;
+	private HashMap<String, UpdatableMetricResult> mMapPrevMetricsResult;
+	private List<SingleValueMetric> metricsOptimized;
```
MichaelRoeder: Why is this separation between "optimized" metrics and the other list of metrics necessary?

src/main/java/org/aksw/simba/lemming/metrics/single/edgemanipulation/VertexDegrees.java
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	/**
+	 * Returns in degree for input vertex id.
+	 * @param vertexId
+	 * @return
+	 */
+	public int getVertexIndegree(int vertexId) {
+		int inDegree = mMapVerticesinDegree[vertexId];
+		return inDegree;
```
MichaelRoeder: Why is the local variable necessary? Wouldn't this be sufficient?
```java
return mMapVerticesinDegree[vertexId];
```

src/main/java/org/aksw/simba/lemming/metrics/single/StdDevVertexDegree.java
```diff
@@ -35,4 +35,51 @@ protected double calculateStdDev(IntArrayList degrees, double avg) {
...
+     * @param graph - input graph.
+     * @param graphOperation - boolean value indicating graph operation. ("true" for adding an edge and "false" for removing an edge). Not used here.
+     * @param triple - Edge on which graph operation is performed. Not used here.
+     * @param newMetricResult - UpdatableMetricResult object containing the results that should be updated.
+     * @return SimpleMetricResult object.
+     */
+    @Override
+    public SimpleMetricResult applyUpdatable(ColouredGraph graph, boolean graphOperation, TripleBaseSingleID triple, UpdatableMetricResult newMetricResult) {
+    	SimpleMetricResult metricResultTempObj = (SimpleMetricResult) newMetricResult;
+    	return metricResultTempObj;
```
MichaelRoeder: This implementation is wrong. The given, previous result is simply returned. Please have a look at the default implementation in the interface.

src/main/java/org/aksw/simba/lemming/metrics/single/MaxVertexDegreeMetric.java
```diff
@@ -28,5 +34,186 @@ public double apply(ColouredGraph graph) {
...
+    /** Stores the previously computed values in UpdateMetricResult object and checks if degrees is not updated for vertices in candidate set.
+     * @param graph - input graph.
+     * @param graphOperation - boolean value indicating graph operation. ("true" for adding an edge and "false" for removing an edge)
+     * @param triple - Edge on which graph operation is performed.
+     * @param newMetricResult - UpdatableMetricResult object containing the results that should be updated.
+     * @return UpdatableMetricResult object with updated candidate set and metric values.
+     */
+    @Override
+    public UpdatableMetricResult applyUpdatable(ColouredGraph graph, boolean graphOperation, TripleBaseSingleID triple, UpdatableMetricResult newMetricResult) {
+    	MaxVertexDegreeMetricResult metricResultTempObj = (MaxVertexDegreeMetricResult) newMetricResult;
```
MichaelRoeder: Reusing the given object like that is not a good idea. It would simply override the results of previous runs and, hence, may change the metric results of other graphs as well.

src/main/java/org/aksw/simba/lemming/metrics/single/MaxVertexDegreeMetric.java
```diff
@@ -28,5 +34,186 @@ public double apply(ColouredGraph graph) {
...
+    	if(graphOperation) {
+    		metricResultTempObj.setCandidatesMetricAddAnEdge(metricResultTempObj.getmMapCandidatesMetricTemp(),metricResultTempObj.getmMapCandidatesMetricValuesTemp());
+    		metricResultTempObj.verifyCandidates(GRAPHOPERATION.AddAnEdgeIndegree, triple);
+    		metricResultTempObj.verifyCandidates(GRAPHOPERATION.AddAnEdgeOutdegree, triple);
+    	}else {
+    		metricResultTempObj.setCandidatesMetricRemoveAnEdge(metricResultTempObj.getmMapCandidatesMetricTemp(),metricResultTempObj.getmMapCandidatesMetricValuesTemp());
+    		metricResultTempObj.verifyCandidates(GRAPHOPERATION.RemoveAnEdgeIndegree, triple);
+    		metricResultTempObj.verifyCandidates(GRAPHOPERATION.RemoveAnEdgeOutdegree, triple);
```
MichaelRoeder: I do not understand why all the functionality is moved into the result object. The metric functionality should be implemented in the metric class. The result object should only contain the result and (maybe) some data that is necessary to create an updated result later on.

src/main/java/org/aksw/simba/lemming/metrics/single/MaxVertexDegreeMetric.java
```diff
@@ -28,5 +34,186 @@ public double apply(ColouredGraph graph) {
...
+			metricResultTempObj.getmMapCandidatesMetricValuesTemp().replace(metricName, metVal); // Store the metric value for later use
+			metricResultTempObj.getmMapCandidatesMetricValues().put(metricName, metVal);
+			
+			IntSet maxDegreeVertices;
+			maxDegreeVertices = mVertexDegrees.getVerticesForDegree((int) metVal, direction); // Get the vertex with the metric value
+			intSetTemp.addAll(maxDegreeVertices); // store the vertex with metric value in candidate set
+
+		} else {
+
+			if (intSetTemp.contains(tripleID)) { // The Edge for vertex in candidate set is modified
```
MichaelRoeder: The set contains vertex IDs while the triple ID is the ID of an edge.

src/main/java/org/aksw/simba/lemming/metrics/single/MaxVertexDegreeMetric.java
```diff
@@ -28,5 +34,186 @@ public double apply(ColouredGraph graph) {
...
+						maxDegreeVertices = mVertexDegrees.getVerticesForDegree((int) metVal, direction); // Get the vertex with the metric value
+						
+						// Remove previous candidates
+						metricResultTempObj.getmMapCandidatesMetricTemp().replace(metricName, new IntOpenHashSet());
+						
+						// store the vertex with metric value in candidate set
+						metricResultTempObj.getmMapCandidatesMetricTemp().replace(metricName, maxDegreeVertices);
+						
+					}
+					metricResultTempObj.getmMapCandidatesMetricValuesTemp().replace(metricName, metVal); // Store the metric value for later use
```
MichaelRoeder: Calling the `replace` method several times with the same metricName does not make sense to me. It is not possible to replace something several times. It will be simply replaced once.

MichaelRoeder (changes requested):
The changes go into the right direction. However, several changes are needed.

First, please ensure to stick to the formatting of the project. To this end, I would suggest to copy the `.editorconfig` file from https://github.com/dice-group/orca/blob/master/.editorconfig and to configure your IDE to work with it.

Second, I doubt that the vertex degree implementation works as it should. It is very complicated while it can be implemented very easily. For example, the Maps in the result object shouldn't be necessary. Apart from that, it seems to contain several flaws. Please have a look at the comments that I left.

src/main/java/org/aksw/simba/lemming/metrics/single/StdDevVertexDegree.java
Hollowfier: The idea behind just passing the previous result in applyUpdatable is to avoid the apply method from being called again.

In the EdgeModifier class, we already calculate the metric value in tryToAddAnEdge/tryToRemoveAnEdge method, 
we thought to reuse it again while calling executeAddingAnEgde/execteRemovingAnEdge.

We are thinking about a better way to implement this.

src/main/java/org/aksw/simba/lemming/metrics/single/edgemanipulation/EdgeModification.java
atulpundir88: I have implemented the suggested changes.

src/main/java/org/aksw/simba/lemming/metrics/single/edgemanipulation/EdgeModifier.java
atulpundir88: For metrics present in the metricsOptimized list, the applyUpdatable method is invoked.  This method saves the result object of different metrics computed by the update method. This result object can be for remove an edge or add an edge depending on the error score. To ensure that the applyUpdatable method is invoked for the metrics which are optimized, we are using metricsOptimized list. 
This rough execution flow is also illustrated in the attached diagram.
[Execution flow.pdf](https://github.com/dice-group/Lemming/files/6642306/Execution.flow.pdf)

src/main/java/org/aksw/simba/lemming/metrics/single/edgemanipulation/VertexDegrees.java
atulpundir88: Yes, I have implemented the suggested change.

src/main/java/org/aksw/simba/lemming/metrics/single/MaxVertexDegreeMetric.java
atulpundir88: The applyUpdatable metric, in this class, will be called for MaxVertexDegree metric. Hence, the previous result object will be of type MaxVertexDegreeMetricResult. Therefore, I have defined it in this manner. So that it can call methods defined specifically for max vertex degree metric result.

src/main/java/org/aksw/simba/lemming/metrics/single/MaxVertexDegreeMetric.java
atulpundir88: I have included the implementation of the verifyCandidates method in MaxVertexDegreeMetric class.

src/main/java/org/aksw/simba/lemming/metrics/single/MaxVertexDegreeMetric.java
atulpundir88: The tripleID variable contains the vertex id, which is passed as an argument to the metricComputationMaxDegree method. I have updated the variable name to vertexID.

src/main/java/org/aksw/simba/lemming/metrics/single/MaxVertexDegreeMetric.java
atulpundir88: I have removed multiple calls to the replace method.

src/main/java/org/aksw/simba/lemming/metrics/single/AvgVertexDegreeMetricResult.java
```diff
@@ -0,0 +1,56 @@
+package org.aksw.simba.lemming.metrics.single;
+
+public class AvgVertexDegreeMetricResult extends SimpleMetricResult {
```
MichaelRoeder, line 11: This class does not seem to be used :confused: 

I do not understand why the class has results for "addEdge" and "removeEdge". This does not fit to the overall workflow which foresees a single result object for each of these cases instead of one result object that stores everything.

src/main/java/org/aksw/simba/lemming/metrics/single/MaxVertexDegreeMetric.java
MichaelRoeder: You misunderstand my argument. I do not say that you may end up with a result object that has the wrong type. My argument is, that you reuse an existing result and override it. This will lead to problems because you delete the results of other graphs and override them with your results!

src/main/java/org/aksw/simba/lemming/metrics/single/edgemanipulation/EdgeModifier.java
MichaelRoeder: Please note that the API of the metrics is created in a way that you _do not have to distinguish_ between the "optimized" metrics and the "normal" metrics. Hence, the list is not necessary.

src/main/java/org/aksw/simba/lemming/metrics/single/StdDevVertexDegree.java
```diff
@@ -8,78 +8,88 @@
...
+		SimpleMetricResult newMetricResult = (SimpleMetricResult) previousResult;
+		int[] degreesArray;
+		IntArrayList degreesList;
 		switch (getName()) {
-			case "stdDevInDegree":	
-				degreesArray = mVertexDegrees.getmMapVerticesinDegree();
-				degreesList =  new IntArrayList(degreesArray);// Arrays.asList(ArrayUtils.toObject(degreesArray));
-				newMetricResult.setResult(calculateStdDev(degreesList, calculateAvg(degreesList)));
-				break;
+		case "stdDevInDegree":
```
MichaelRoeder: String comparisons are costly. Using `this.direction` is much better.

src/main/java/org/aksw/simba/lemming/metrics/single/StdDevVertexDegree.java
```diff
@@ -8,78 +8,88 @@
...
+			degreesArray = mVertexDegrees.getmMapVerticesinDegree();
```
MichaelRoeder: The name of the method `getmMapVerticesinDegree` does not use correct CamelCase.

src/main/java/org/aksw/simba/lemming/metrics/single/StdDevVertexDegree.java
```diff
@@ -8,78 +8,88 @@
...
+			degreesList = new IntArrayList(degreesArray);// Arrays.asList(ArrayUtils.toObject(degreesArray));
+			newMetricResult.setResult(calculateStdDev(degreesList, calculateAvg(degreesList)));
+			break;
+
+		case "stdDevOutDegree":
+			degreesArray = mVertexDegrees.getmMapVerticesoutDegree();
```
MichaelRoeder: The name of the method `getmMapVerticesoutDegree` does not use correct CamelCase.

MichaelRoeder (changes requested):
Again: ensure to stick to the formatting of the project. The current formatting is still wrong.

I added further comments to single issues.

Please think about a further optimization. At the moment, the optimizations that are applied are not very effective. So please think about which information is really necessary to calculate the standard deviation.

MichaelRoeder (changes requested):
I write it a third and last time: **please ensure to stick to the formatting of the project**. The current formatting is wrong and I have to stop reviewing this PR as long as the formatting is not correct as it marks a lot of changes that are simply changes of the formatting but not changes of the content.

src/main/java/org/aksw/simba/lemming/metrics/single/edgemanipulation/VertexDegrees.java
```diff
@@ -0,0 +1,181 @@
...
+import it.unimi.dsi.fastutil.ints.IntOpenHashSet;
+import it.unimi.dsi.fastutil.ints.IntSet;
+
+/**
+ * Class for storing in and out degree of vertices.
+ * 
+ * @author Atul
+ *
+ */
+public class VertexDegrees {
```
MichaelRoeder, line 16: @atulpundir88 I would like you to talk with @wangzun66 about whether it would make sense to remove the vertex degree class and replace it with a class that stores a histogram of the in and out degrees.

Wang Zun reported that he already tried it but was not sure whether it would use too much memory. However, I think it might be a better alternative than this class. :thinking:

PR #25, opened 2021-09-05
optimize triangle metric

wangzun66:


MichaelRoeder (commented):
Apart from the discussion we had in the meeting (we talked about whether a "loop" edge can be part of an edge triangle or not), I would like to point out that it might be better to make the triangles implement the `update` method we discussed for the other metrics. This would allow you to move the code for updating the number of triangles from the EdgeModification class into the triangle class implementations (where it actually belongs to).

wangzun66:
@MichaelRoeder For the loop edge part that we discussed, I didn't find the exact definitions of node triangle and edge triangle. But I found a definition of triangle in a graph, it is from the thesis "Finding, Counting and Listing all Triangles in Large Graphs, An Experimental Study".

截屏2021-10-23 19 03 47

So I think, two nodes can not form a triangle, no matter node triangle or edge triangle.

But I am also confused with another words in the thesis: 

截屏2021-10-23 19 08 21

According to these words, two nodes with a loop edge should be seen as a node/edge triangle.
Can you give me a suggestion?

wangzun66:
@MichaelRoeder And I want to confirm a point from the " <code>update</code> method part". Should I delete EdgeModification after I moving the code into triangle metrics?? Because this class is only used to update triangle metics, and it stores also an instance of <code>Grph</code>, wastes a lot of storage.

PR #27, opened 2021-10-31
Metrics optimization

wangzun66:
This is a merge branch for all metrics' optimization!


src/main/java/org/aksw/simba/lemming/metrics/MetricUtils.java
```diff
@@ -47,10 +50,32 @@
...
+        neighborsOfConnectedVertices[1] = graph.getInNeighbors(v2);
+        neighborsOfConnectedVertices[1].addAll(graph.getOutNeighbors(v2));
+
+        if (neighborsOfConnectedVertices[1].contains(v1))
+            neighborsOfConnectedVertices[1].remove(v1);
+        if (neighborsOfConnectedVertices[1].contains(v2))
+            neighborsOfConnectedVertices[1].remove(v2);
+
+        return IntSetUtil.intersection(neighborsOfConnectedVertices[0], neighborsOfConnectedVertices[1]);
+    }
```
MichaelRoeder: 1. Javadoc is missing. It is unclear for the user why v1 and v2 are removed if you do not explicitly document it.
2. What about unit tests for this method?
3. Wouldn't the following be easier?
```java
    public static IntSet getVerticesInCommon(ColouredGraph graph, int v1, int v2) {
        IntSet v1Neighbors;
        v1Neighbors = graph.getInNeighbors(v1);
        v1Neighbors.addAll(graph.getOutNeighbors(v1));
        
        IntSet intersection = IntSetUtil.intersection(v1Neighbors, graph.getInNeighbors(v2));
        intersection.addAll(IntSetUtil.intersection(v1Neighbors, graph.getOutNeighbors(v2)));

        if (intersection.contains(v1))
            intersection.remove(v1);
        if (intersection.contains(v2))
            intersection.remove(v2);
        
        return intersection;
    }
```
It avoids the creation of the set that contains all neighbors of v2 since it is not needed to create it. It also removes `v1` and `v2` only from the result instead of removing it from intermediate results.

src/main/java/org/aksw/simba/lemming/metrics/single/MaxVertexDegreeMetric.java
```diff
@@ -29,4 +34,55 @@ public double apply(ColouredGraph graph) {
...
+        // variable to track remove an edge or add an edge operation.
+
+        double metVal = ((SingleValueMetricResult) previousResult).getResult();
+        int changedDegree = getChangedDegree(graph, vertexID, direction);
+        int degree = changedDegree - updateVertexDegree;
+        if (updateVertexDegree == -1) { // Remove an edge
+            if (degree == metVal) {// If degree of a vertex is equal to maximum vertex then apply method is called.
+                metVal = apply(graph);
+            }
+        } else { // Add an edge
```
MichaelRoeder: I thought that this metric also stores the number of nodes that have the max vertex degree so that it only has to count again, if there is no other vertex that still has the maximum degree. What happened to this idea? :confused:

src/main/java/org/aksw/simba/lemming/metrics/single/SingleValueMetricResult.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * The class stores the metric name and the result which can be reused.
+ * 
+ * @author Atul
+ *
+ */
+public class SingleValueMetricResult implements UpdatableMetricResult {
+
+    protected String metricName;
+    protected double result;
```
MichaelRoeder: Javadoc comments are missing :wink:

src/main/java/org/aksw/simba/lemming/metrics/single/StdDevVertexDegree.java
```diff
@@ -35,4 +38,105 @@ protected double calculateStdDev(IntArrayList degrees, double avg) {
...
+     * @param avg              - average of previous set of nodes in the graph
+     * @param variance         - variance of previous set of nodes
+     * @param oldDegree        - the degree which was updated after adding or
+     *                         removing an edge
+     * @param graphOperation   - denotes if an edge was added or removed
+     * @return List<Double> - a list containing average and variance in that order.
+     */
+    private List<Double> computeAvgVarianceFromPreviousResult(double numberOfVertices, double avg, double variance,
+            double newDegree, Operation graphOperation) {
+        List<Double> list = new ArrayList<Double>();
```
MichaelRoeder: A list of `Double` values? That is pretty expensive to return two simple numbers...

Why not `double result[] = new double[];`? It allows to return two values, uses less memory than an ArrayList and avoids the conversion of the double values into their `Double` objects.

src/main/java/org/aksw/simba/lemming/metrics/single/StdDevVertexDegree.java
```diff
@@ -35,4 +38,105 @@ protected double calculateStdDev(IntArrayList degrees, double avg) {
...
+        double flag = graphOperation == Operation.ADD ? 1 : -1;
```
MichaelRoeder: `flag` is not a good name for this variable. Please rethink your choice :wink:

src/main/java/org/aksw/simba/lemming/metrics/single/StdDevVertexDegreeMetricResult.java
```diff
@@ -0,0 +1,87 @@
...
+     * 
+     * @author Pranav
+     *
+     */
+
+    private double varianceVertexDegrees = 0.0; // Variable for storing variance
+
+    private double avgVertexDegrees = 0.0; // Variable for storing average of vertices.
+
+    private double numberOfVertices = 0.0; // Variable for storing number of vertices.
```
MichaelRoeder: Nice comments. Why not as Javadoc so that a user can access them in an easier way? :wink:

src/main/java/org/aksw/simba/lemming/metrics/single/StdDevVertexDegreeMetricResult.java
```diff
@@ -0,0 +1,87 @@
...
+
+public class StdDevVertexDegreeMetricResult extends SingleValueMetricResult {
+    /**
+     * The class stores the average, variance and the number of vertices, which is
+     * used for computing the standard deviation vertex degrees metric in multiple
+     * iterations.
+     * 
+     * @author Pranav
+     *
+     */
```
MichaelRoeder: An interesting Javadoc comment. Unfortunately, it does not work as it should because it has to be put directly _in front_ of the class.

src/main/java/org/aksw/simba/lemming/metrics/single/StdDevVertexDegreeMetricResult.java
```diff
@@ -0,0 +1,87 @@
...
+     * @param metricName - Name of the metric.
+     * @param result     - Metric value.
+     */
+    public StdDevVertexDegreeMetricResult(String metricName, double result) {
+        super(metricName, result);
+    }
+
+    /**
+     * 
+     * Returns the variance of in degrees computed after any edge operation
```
MichaelRoeder: "... of in degrees ..." Why only in-degrees? (Applies also to other comments in this class)

src/main/java/org/aksw/simba/lemming/metrics/single/edgetriangles/EdgeTriangleMetric.java
```diff
@@ -16,9 +24,59 @@ public double apply(ColouredGraph graph) {
...
+		if(headId == tailId){
+			return previousResult;
+		}
+
+		int change = opt==Operation.REMOVE ? -1 : 1 ;
+		int numEdgesBetweenVertices = IntSetUtil.intersection(graph.getEdgesIncidentTo(tailId), graph.getEdgesIncidentTo(headId)).size();
+
+		int differenceOfSubGraph = calculateDifferenceOfSubGraphEdge(graph, headId, tailId, numEdgesBetweenVertices, change);
+		double newResult = previousResult.getResult() + change*differenceOfSubGraph;
+		newResult = newResult >= 0 ? newResult: 0;
```
MichaelRoeder: Why this line is here? The new result should never be negative. Otherwise, something went wrong before. Hence, I would suggest to throw an exception or at least to log an error instead of fixing it quietly.

src/main/java/org/aksw/simba/lemming/metrics/single/nodetriangles/NodeTriangleMetric.java
```diff
@@ -17,8 +25,45 @@ public double apply(ColouredGraph graph) {
...
+		if(numEdgesBetweenVertices==1 && opt == Operation.REMOVE){
+			newResult = newResult - numberOfCommon;
+
+		//2.case: add an edge, and number of edges between head and tail is 0
+		// -> new metric = old metric + number of common vertices
+		}else if (numEdgesBetweenVertices==0 && opt == Operation.ADD){
+			newResult = newResult + numberOfCommon;
+		}
+
+		newResult = newResult>=0 ? newResult : 0;
```
MichaelRoeder: The same as for the edge triangle metric.

MichaelRoeder (changes requested):
Please have a look at my comments.

src/main/java/org/aksw/simba/lemming/metrics/MetricUtils.java
wangzun66: Done.

src/main/java/org/aksw/simba/lemming/metrics/single/edgetriangles/EdgeTriangleMetric.java
wangzun66: Done.

src/main/java/org/aksw/simba/lemming/metrics/single/nodetriangles/NodeTriangleMetric.java
wangzun66: Done.

src/main/java/org/aksw/simba/lemming/metrics/single/SingleValueMetricResult.java
pra-kulk: Done

src/main/java/org/aksw/simba/lemming/metrics/single/StdDevVertexDegree.java
pra-kulk: Done

src/main/java/org/aksw/simba/lemming/metrics/single/StdDevVertexDegree.java
pra-kulk: Done

src/main/java/org/aksw/simba/lemming/metrics/single/StdDevVertexDegreeMetricResult.java
pra-kulk: Done

src/main/java/org/aksw/simba/lemming/metrics/single/StdDevVertexDegreeMetricResult.java
pra-kulk: Done

src/main/java/org/aksw/simba/lemming/metrics/single/StdDevVertexDegreeMetricResult.java
pra-kulk: Done

src/main/java/org/aksw/simba/lemming/metrics/single/MaxVertexDegreeMetric.java
atulpundir88: Done. This logic is included now.

PR #29, opened 2021-11-21
Parallel processing

Hollowfier:
- Multithreading without synchronized block
- Added ColouredGraphDecorator class
- AddEdgeDecorator and RemoveEdgeDecorator to _"predict"_ the metric values on Edge operation
- Changed Execution flow (Now the graph is not reverted to original version again and again after adding/removing an edge)


src/main/java/org/aksw/simba/lemming/ColouredGraphDecorator.java
```diff
@@ -0,0 +1,300 @@
...
+
+    /**
+     * Get number of edges between two vertices
+     * 
+     * @return int - number of edges
+     */
+    public int getNumberOfEdgesBetweenVertices() {
+        return IntSetUtil.intersection(this.graph.getEdgesIncidentTo(this.triple.tailId),
+                this.graph.getEdgesIncidentTo(this.triple.headId)).size();
+    }
```
MichaelRoeder: 1. From the javadoc comment, I do not get the meaning of this method. Is it the number of edges between any two vertices or between two specific vertices?
2. The implementation is not really efficient. It creates a new IntSet just to recognize the size of it. After that, it will throw the set away. That can be done much better :wink:

src/main/java/org/aksw/simba/lemming/ColouredGraphDecorator.java
```diff
@@ -0,0 +1,300 @@
...
+
+    protected IColouredGraph graph;
+    protected boolean isAddingEdge;
+    protected TripleBaseSingleID triple;
+
+    public ColouredGraphDecorator() {
+        this.graph = null;
+        this.isAddingEdge = false;
+        this.triple = null;
+    }
```
MichaelRoeder: Does a decorator that does not have anything to decorate really make sense?

src/main/java/org/aksw/simba/lemming/ColouredGraphDecorator.java
```diff
@@ -0,0 +1,300 @@
...
+/**
+ * Base Decorator for Coloured Graph Class
+ * 
+ * @author Pranav
+ */
+public class ColouredGraphDecorator implements IColouredGraph {
+
+    protected IColouredGraph graph;
+    protected boolean isAddingEdge;
+    protected TripleBaseSingleID triple;
```
MichaelRoeder: Are these two attributes really a part of a generic decorator? Wouldn't it make more sense to have a generic decorator class that handles the decorator functionality and an additional class that handles the specific case in which somebody wants to change a single edge? :thinking:

src/main/java/org/aksw/simba/lemming/ColouredGraphDecorator.java
```diff
@@ -0,0 +1,300 @@
...
+    /**
+     * Get list of all Edge IDs connecting to vertex
+     *
+     * @param verticeId - verticeId the id of an vertex
+     * @return IntSet - set of edge IDs
+     */
+    @Override
+    public IntSet getEdgesIncidentTo(int verticeId) {
+        return this.graph.getEdgesIncidentTo(verticeId);
+    }
```
MichaelRoeder: Hint: When you override a method of a super class, an "own" javadoc is only necessary if you want to add an important implementation detail. The method will "inherit" the javadoc of the original method :wink:

src/main/java/org/aksw/simba/lemming/ColouredGraphDecorator.java
```diff
@@ -0,0 +1,300 @@
...
+import it.unimi.dsi.fastutil.ints.IntSet;
+
+/**
+ * Base Decorator for Coloured Graph Class
+ * 
+ * @author Pranav
+ */
+public class ColouredGraphDecorator implements IColouredGraph {
+
+    protected IColouredGraph graph;
```
MichaelRoeder: We have a lot of variables that are just named `graph` :sweat_smile: 
Maybe `decorated` or `decoratedGraph` would be more precise :wink:

src/main/java/org/aksw/simba/lemming/IColouredGraph.java
```diff
@@ -0,0 +1,171 @@
...
+import com.carrotsearch.hppc.BitSet;
+
+import it.unimi.dsi.fastutil.ints.IntArrayList;
+import it.unimi.dsi.fastutil.ints.IntSet;
+
+/**
+ * This interface is used for the decorator pattern to simplify multithreading
+ * in the graph amendment phase
+ * 
+ * @author Pranav
```
MichaelRoeder: This is not really a good description of the interface and its usage :wink:

src/main/java/org/aksw/simba/lemming/RemoveEdgeDecorator.java
```diff
@@ -0,0 +1,198 @@
...
+ */
+package org.aksw.simba.lemming;
+
+import it.unimi.dsi.fastutil.ints.IntArrayList;
+import it.unimi.dsi.fastutil.ints.IntSet;
+
+/**
+ * 
+ * Concrete Decorator class to simulate removing of an edge to the ColouredGraph
+ * object in the current iteration
```
MichaelRoeder: "... to simulate the _removal_ of an edge _from_ the _IColouredGraph_ object."

src/main/java/org/aksw/simba/lemming/RemoveEdgeDecorator.java
```diff
@@ -0,0 +1,198 @@
...
+ *
+ */
+public class RemoveEdgeDecorator extends ColouredGraphDecorator {
+
+    /**
+     * 
+     */
+    public RemoveEdgeDecorator() {
+        super();
+    }
```
MichaelRoeder: Again, a decorator without anything to decorate typically does not make much sense :thinking:

src/main/java/org/aksw/simba/lemming/creation/GeologyDataset.java
```diff
@@ -27,41 +27,27 @@ public GeologyDataset() {
...
-			ontModel.read("geology/sosa.ttl");
-			ontModel.read("geology/thors.ttl");
-			ontModel.read("geology/time.ttl");
-			ontModel.read("geology/basic.ttl");
-			ontModel.read("geology/temporal.ttl");
-			
+			File ontFolder = new File("geology");
+			for(File file : ontFolder.listFiles()){
+				ontModel.read(file.getAbsolutePath(), "ttl");
+			}
```
MichaelRoeder: Nice! :+1:

MichaelRoeder (commented):
I added some first comments but didn't went through all files. However, you can already have a look at the comments.

src/main/java/org/aksw/simba/lemming/ColouredGraphDecorator.java
pra-kulk: Improved the implementaion and the javadoc.

src/main/java/org/aksw/simba/lemming/ColouredGraphDecorator.java
pra-kulk: The two attributes are not used in this base decorator. However, they are needed in the AddEdgeDecorator and RemoveEdgeDecorator classes which inherit these attributes and their get(), set() methods. 

Removing them from here would mean they would be added to both these sub classes. I just included them here to not to repeat the code in both sub classes.

src/main/java/org/aksw/simba/lemming/ColouredGraphDecorator.java
pra-kulk: Changed.

src/main/java/org/aksw/simba/lemming/ColouredGraphDecorator.java
pra-kulk: Removed.

src/main/java/org/aksw/simba/lemming/IColouredGraph.java
pra-kulk: Improved the javadoc

src/main/java/org/aksw/simba/lemming/RemoveEdgeDecorator.java
pra-kulk: Changed

src/main/java/org/aksw/simba/lemming/RemoveEdgeDecorator.java
pra-kulk: Done

src/main/java/org/aksw/simba/lemming/ColouredGraphDecorator.java
MichaelRoeder: Fair point. Typically, I would argue to add another class in-between that offers this functionality. However, I think we can go with your solution as well :+1:

src/main/java/org/aksw/simba/lemming/AddEdgeDecorator.java
```diff
@@ -0,0 +1,191 @@
...
+
+    /**
+     * Get number of edges between two vertices after given edge has been added to
+     * them
+     * 
+     * @return int - number of edges
+     */
+    @Override
+    public int getNumberOfEdgesBetweenVertices(int tailId, int headId) {
+        return super.getNumberOfEdgesBetweenVertices(tailId, headId) + 1;
```
MichaelRoeder: Don't you have to check whether the given head and tailIDs are the IDs of the two nodes that have the added edge?  :thinking: 

Remember: this method is public. Everybody could call it with any IDs :wink:

src/main/java/org/aksw/simba/lemming/RemoveEdgeDecorator.java
```diff
@@ -0,0 +1,192 @@
...
+
+    /**
+     * Get number of edges between two vertices after given edge has been added to
+     * them
+     * 
+     * @return int - number of edges
+     */
+    @Override
+    public int getNumberOfEdgesBetweenVertices(int tailId, int headId) {
+        return super.getNumberOfEdgesBetweenVertices(tailId, headId) - 1;
```
MichaelRoeder: The same as in the other decorator: Don't you have to check whether the given head and tail IDs are the IDs of the two nodes that have the removed edge?  :thinking: 

Remember: this method is public. Everybody could call it with any IDs :wink:

MichaelRoeder (changes requested):
I think only the issue with the two commented methods remains.

src/main/java/org/aksw/simba/lemming/AddEdgeDecorator.java
Hollowfier: Fixed it now. 

> Everybody could call it with any IDs

Currently I log an error onto the console and return -1 if any of the two node Ids is not a part of the added edge. 
Let me know if this is correct. If not, in what other way can I handle this case ?

src/main/java/org/aksw/simba/lemming/RemoveEdgeDecorator.java
Hollowfier: Same as other decorator.

> Everybody could call it with any IDs

Currently I log an error onto the console and return -1 if any of the two node Ids is not a part of the added edge.
Let me know if this is correct. If not, in what other way can I handle this case ?

PR #32, opened 2021-12-04
Compute only metrics present in Characteristics Expressions

atulpundir88:
Created a method in ConstantValueStorage class. This method as an input takes a list of metrics. The list will contain all the metrics that are configured for the Lemming project. The method checks which metrics are part of the expression and returns an updated list that includes only metrics that are present in the expressions.

### DIFF
--- a/src/main/java/org/aksw/simba/lemming/mimicgraph/metricstorage/ConstantValueStorage.java
+++ b/src/main/java/org/aksw/simba/lemming/mimicgraph/metricstorage/ConstantValueStorage.java
@@ -239,7 +239,7 @@ public class ConstantValueStorage implements Serializable	{
 	 * @param lstMetrics - The list which contains all the input metrics.
 	 * @return List of metrics
 	 */
-	public List<SingleValueMetric> getMetrics(List<SingleValueMetric> lstMetrics){
+	public List<SingleValueMetric> getMetricsOfExpressions(List<SingleValueMetric> lstMetrics){
 	    List<SingleValueMetric> metrics = new ArrayList<>();// List which will contain metrics present in Expressions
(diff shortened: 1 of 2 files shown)


src/main/java/org/aksw/simba/lemming/mimicgraph/metricstorage/ConstantValueStorage.java
```diff
@@ -232,4 +234,34 @@ public void loadData(){
 	public static String generateGraphKey(ColouredGraph graph) {
 	    return graph.getGraph().getNumberOfVertices() + "-" + graph.getGraph().getNumberOfEdges();
 	}
+	
+	/** The method returns the list of metrics which are present in characteristics expressions.
+	 * @param lstMetrics - The list which contains all the input metrics.
+	 * @return List of metrics
+	 */
+	public List<SingleValueMetric> getMetrics(List<SingleValueMetric> lstMetrics){
```
MichaelRoeder: The name of the method is not really expressive. `getX` typically means that the class knows X and gives it to you. However, in this case, you already have the list that includes X and want a shorter, filtered list. So maybe `filterMetrics` or `getMetricsOfExpressions` would be better. :thinking:

MichaelRoeder (commented):
The method name is misleading (from my point of view). Apart from that, the PR looks good.

src/main/java/org/aksw/simba/lemming/mimicgraph/metricstorage/ConstantValueStorage.java
atulpundir88: Updated the method name to getMetricsOfExpressions.

